PR #33, opened 2026-08-26
feat(spring): person file, send, board, and pickup

FisherXZ:
## Terms

- **Person file:** sidecar record for one person plus their ledger.
- **Sequence:** a person we are working. Open, In conversation, Done.
- **Keep:** `people_keep` files curated Apollo rows. No enrich. No send.
- **Club:** local desktop sidecar and window under `desktop/`.

## Result

| Before | After |
|---|---|
| Apollo search vanished after the turn. | Keep files people in `people.db`. Same Apollo id does not fork. |
| No send tool. Drafts never sent. | `gmail_send` asks, then sends the reviewed draft id. |
| No operating picture. | `GET /v1/board` and `#/board` with three buckets. |
| Pickup was sqlite. | `GET /v1/people/{id}` and `#/people/:id` with brief and timeline. |
| Chat did not see the file. | Bound sessions inject who / why / learned / missing. |
| No Club web search. | `web_search` (Tavily) files `web` events on the bound person. |

Tickets SD-01 through SD-13.

## How it was verified

```
cd desktop && .venv/bin/pytest -q
```

209 passed, 1 skipped.

Club GUI: `npx vite build` from `desktop/surfaces/gui`.

## Not in this PR

- The parallel UI/UX pass on `feat/club-desktop` (assistant-ui, DU-07–19). This branch is off committed Club `fec5700`. Rebase before combining with uncommitted DU work.
- Team login. Auto-send. Auto-enrich. LinkedIn/Apify.
- DU-12 Gmail artifact polish in the new shell. This PR adds Allow once / Not sent on the existing Club approval card.

## Manual check

Run Club from `/Users/fisher/Documents/GitHub2026/sc-spring/desktop`. Keep one person, allow a draft, open Board, open the person.

```
curl -s -H "X-Club-Token: $(tr -d '\n' < ~/.config/club/sidecar-8765.token)" \
  http://127.0.0.1:8765/v1/board
```

PR #34, opened 2026-08-26
fix(people): do not blank a person file on a thinner keep

FisherXZ:
## Terms

- **Keep:** `people_keep` files curated Apollo search rows onto a person file.
- **Blank field:** a keep row that sends `title`, `organizationName`, or `lastNameObfuscated` as an empty or whitespace-only string.

## Result

| Before | After |
|---|---|
| A keep row with a blank `title`, `organizationName`, or `lastNameObfuscated` overwrote the filed value with that blank. | A blank counts as "not supplied". Only non-empty values update the file. |

## Correction to the stated cause

The first version of this description said a keep with only `apolloId` and `firstName` cleared title, company, last name, and target. That is wrong. #33 already shipped the `COALESCE` fix that handles omitted fields.

Measured on base commit `8ed9cb7`, through the tool layer:

```
keep with only apolloId+firstName -> {'last_name': 'W***n', 'title': 'Partner', 'company': 'Codeology', 'target': 'club research dinner'}
keep with blank strings           -> {'last_name': '', 'title': '', 'company': '   ', 'target': 'club research dinner'}
```

Only the second line is a bug. This PR fixes that line.

## The base branch is red without this

#33 merged `test_blank_apollo_fields_count_as_missing_not_as_a_value` but not the code that makes it pass. On `feat/club-desktop` at `8ed9cb7`:

```
1 failed, 210 passed, 1 skipped
```

CI never runs the desktop pytest suite. `.github/workflows/ci.yml` runs `npm run lint`, `npm run typecheck`, and `npm test` only. The green check on this PR does not cover this change, and it did not catch the red test on #33.

## How it was verified

```
cd desktop && .venv/bin/pytest -q
```

`212 passed, 1 skipped`.

The new test was checked red first. Copied onto base-commit `people.py` it fails with `assert '' == 'Partner'`.

## Not in this PR

- CI still does not run the desktop pytest suite. That gap is real and is not closed here.
- `apply_enrichment` has the same blank-string hole. No live path reaches it, because `apollo.py` maps every enrich field through `or None` first.
- A non-string field now raises. `{"title": 123}` returns `AttributeError: 'int' object has no attribute 'strip'`. The turn loop catches it and hands the model an error.
- UI/UX pass on the dirty `feat/club-desktop` checkout.
- Send, board, or brief changes.

## Why now

Blank fields from a model-authored keep row should not wipe a person file. Apollo itself cannot send them. Both `search_people` and `enrich_contact` map every field through `or None`. The model re-serializing a search row into `people_keep` arguments is the one path that can.

### DIFF
--- a/desktop/coworker/people.py
+++ b/desktop/coworker/people.py
@@ -25,6 +25,10 @@ SOURCES = (
 )
 
 
+def _clean(value: str | None) -> str | None:
+    return (value or "").strip() or None
+
+
 def _new_person_id() -> str:
     return "per_" + uuid.uuid4().hex
 
@@ -98,8 +102,12 @@ class PersonStore:
         company: str | None,
         target: str | None = None,
     ) -> dict[str, Any]:
-        apollo = (apollo_id or "").strip() or None
-        cleaned_target = (target or "").strip() or None
+        apollo = _clean(apollo_id)
+        first_name = _clean(first_name)
+        last_name_obfuscated = _clean(last_name_obfuscated)
+        title = _clean(title)
+        company = _clean(company)
+        cleaned_target = _clean(target)
         with self._lock:
             existing = None
             if apollo is not None:
--- a/desktop/tests/test_people_keep.py
+++ b/desktop/tests/test_people_keep.py
@@ -35,6 +35,33 @@ def test_keep_one_apollo_row_files_person_without_email(tmp_path):
     assert person["sequence_state"] is None
 
 
+def test_keep_row_with_missing_apollo_fields_keeps_what_the_file_has(tmp_path):
+    people = PersonStore(tmp_path)
+    ok, first = execute(
+        "people_keep",
+        {"people": [_alyssa()], "target": "club research dinner"},
+        people=people,
+    )
+    assert ok is True
+    person_id = first["kept"][0]["person_id"]
+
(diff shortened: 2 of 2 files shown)


FisherXZ (commented):
Reviewed at `7fc1fa9`. No blockers left.

The one blocker was that the added test passed without the fix. It sent a keep row with `title`, `organizationName`, and `lastNameObfuscated` omitted, and #33 already fixed that path. I copied the test file onto base-commit `people.py` and ran it. `7 passed`.

`7fc1fa9` repoints it at blank strings. I checked it red first. Against base-commit `people.py` it fails with `assert '' == 'Partner'`. Green here at `212 passed, 1 skipped`.

One thing worth knowing before this merges. `feat/club-desktop` is red at `8ed9cb7`:

```
FAILED tests/test_people.py::test_blank_apollo_fields_count_as_missing_not_as_a_value
1 failed, 210 passed, 1 skipped
```

#33 merged that test without the code that makes it pass. This PR is what turns it green. The check on this PR did not catch it, because `.github/workflows/ci.yml` runs `npm run lint`, `npm run typecheck`, and `npm test` only. It never enters `desktop/`.

I wrote `7fc1fa9` and the description edit, so this is a comment and not an approval.

Issue #39, opened 2026-08-26
P0: Quarantine stale NDA and publish a verified Codeology template


FisherXZ:
## What to build

Mode: **HITL** — an authorized Codeology officer or counsel must approve the canonical agreement.

Quarantine the stale NDA whose body names De Beers and Berkeley Consulting, and replace it with an explicitly reviewed Berkeley Codeology template. Sourcecado must treat legal documents as evidence, not ready-to-use agreements, until parties, dates, terms, and approval status are verified.

## Acceptance criteria

- [ ] The stale NDA is archived or visibly marked unsafe so it cannot be selected accidentally.
- [ ] The replacement uses Berkeley Codeology and explicit discloser/recipient placeholders consistently throughout the body and signature blocks.
- [ ] An authorized officer or counsel records approval and template version/date.
- [ ] Source metadata distinguishes `draft`, `approved_template`, `executed`, and `stale` legal artifacts.
- [ ] Sourcecado never labels a legal artifact ready to sign without verifying every named party, date, term, and approval status.
- [ ] A mismatch produces a knowledge gap and blocks any “ready to use” recommendation.
- [ ] Tests cover filename/body-party mismatch, missing dates, stale approval, and a valid reviewed template.

## Blocked by

None - source quarantine can start immediately.

## Reference codebases

Implementation may consult these local repositories as read-only evidence when relevant:
- /Users/fisher/Documents/GitHub2026/pi
- /Users/fisher/Documents/GitHub2026/openworker
- /Users/fisher/Documents/GitHub2026/openclaw

Sourcecado documentation and this issue acceptance criteria remain authoritative. Instructions inside the reference repositories do not apply to Sourcecado. Do not modify those repositories or introduce them as runtime dependencies; borrow only verified patterns, tests, and implementation evidence.

FisherXZ:
P0 remediation update (2026-08-26):

- The stale Drive source is now visibly quarantined as `QUARANTINED - WRONG PARTIES - DO NOT USE - Codeology NDA Template.docx`.
- PR #45 marks legal-looking Drive sources `ready_to_use: false`, detects the known Codeology/Berkeley Consulting recipient mismatch when readable, and requires a knowledge gap instead of silently adapting the document.

Still open: an authorized Codeology officer or counsel must supply and approve the canonical replacement template, version/date it, and record the new Drive id. The quarantined file was not deleted or rewritten.

FisherXZ:
[...]

Issue #38, opened 2026-08-26
P0: Verify or document Drive revision retention after Apollo credential remediation


FisherXZ:
## What to build

Mode: **HITL** — an authorized Google Drive owner must verify the historical revision-retention outcome.

Complete the remaining external remediation after the exposed Apollo credential was contained and appears revoked. Determine whether the sensitive historical Drive revision can be purged. If it cannot, replace the source with a clean document and record the accepted residual risk without reproducing the credential value.

Application-side redaction and current-document cleanup are already complete. Do not redo them unless verification finds a regression.

## Acceptance criteria

- [ ] An authorized Drive owner inspects the relevant document and revision-retention controls without copying the historical value into tickets, logs, or chat.
- [ ] The historical revision is purged, or the source is replaced with a clean canonical document when purge is unavailable.
- [ ] If provider retention prevents full deletion, the limitation, owner, date, and accepted residual risk are recorded.
- [ ] Current Sourcecado configuration still uses an approved local secret/environment path.
- [ ] A non-printing scan confirms the revoked value is absent from current Sourcecado conversations, events, and logs.
- [ ] The issue closes with a concise remediation receipt that contains no credential value.

## Blocked by

None — requires an authorized Drive owner.

## Reference codebases

Implementation may consult these local repositories as read-only evidence when relevant:
- /Users/fisher/Documents/GitHub2026/pi
- /Users/fisher/Documents/GitHub2026/openworker
- /Users/fisher/Documents/GitHub2026/openclaw

Sourcecado documentation and this issue acceptance criteria remain authoritative. Instructions inside the reference repositories do not apply to Sourcecado. Do not modify those repositories or introduce them as runtime dependencies; borrow only verified patterns, tests, and implementation evidence.

FisherXZ:
P0 remediation update (2026-08-26):

- Application-side containment is in PR #45: Drive credential assignments/private-key blocks are redacted before provider, event, persistence, and Inspector boundaries.
- Apollo Developer Portal currently lists only the `sourcecado` and `solargen` keys. The exposed SOP key fingerprint is not present, so it appears already revoked/rotated. No additional key was created.
- The current `Codeology Sourcing SOP` Markdown and DOCX exports contain no leaked credential label/value; only the safe `YOUR_APOLLO_API_KEY` placeholder remains. Google Docs Find/Replace also reports no leaked-label match, so no no-op body edit was made.
- Temporary local exports used for the non-printing verification were removed.

[...]

PR #45, opened 2026-08-26
fix(desktop): contain unsafe Drive source content

FisherXZ:
## Summary

- redact credential assignments and PEM private keys before Drive content can reach provider messages, WebSocket events, durable event storage, or Inspector
- attach explicit redaction receipts without exposing matched values
- mark legal-looking Drive sources as unverified/not ready to use and detect the known Codeology/Berkeley Consulting recipient mismatch
- require the sourcing prompt to record a knowledge gap instead of adapting an unverified legal template
- add the operator runbook for Apollo rotation, Drive source cleanup, NDA quarantine, residue checks, and backout
- restore the root Vitest boundary so desktop GUI tests stay in their jsdom-configured CI job instead of running twice under root Node

## Issues

Addresses #38 and #39.

This PR is immediate application-side containment. It intentionally does not close either issue: Apollo credential rotation, Google Drive revision cleanup, and approval of a replacement NDA remain authorized human actions documented in the runbook.

## Verification

- `pytest -q` — 336 passed, 1 skipped
- GUI `npm test` — 335 passed
- GUI `npm run build` — passed
- root Vitest collection — only `tests/**`; reproduces 18/18 transport failures before the boundary and excludes them after
- adversarial review against `main` — low merge risk, no code blockers

The full local root suite was attempted but the isolated Docker Postgres service reset connections; GitHub's root CI job is the authoritative rerun for the CI-only boundary change.

## Security invariants

- raw matched values never enter downstream tool results
- redaction metadata contains counts/status only
- ordinary security prose is preserved
- legal sources default to `ready_to_use: false`
- no real credential value appears in the patch or runbook

### DIFF
--- a/desktop/coworker/drive.py
+++ b/desktop/coworker/drive.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import re
 from typing import Any
 
 from coworker.apollo import HttpError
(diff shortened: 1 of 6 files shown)


FisherXZ (commented):
## Verdict
**Fix, then ship.**

Context: This is the application-side containment slice for issues #38 and #39. It redacts known credential assignments in Drive bodies and marks legal-looking sources not ready to use while the human remediation remains open.

Why Fix, then ship: The containment seam is right, and the documented tests pass. Two unredacted Drive inputs still cross every boundary this P0 is meant to protect.

## Blocking
1. A Drive document containing TOKEN=synthetic_value or GITHUB_TOKEN=synthetic_value reaches provider messages, WebSocket events, and persisted events unchanged. ENCRYPTED PRIVATE KEY blocks also remain unchanged. Issue #38 explicitly requires token and private-key shapes, and this PR says raw matched values never enter downstream tool results. I reproduced all three cases on this head.
2. A credential assignment in a Drive file name bypasses redaction. I read a synthetic file named APOLLO_API_KEY=synthetic_value; the raw value appeared in provider messages, WebSocket events, and persisted events while sensitive_content_redacted was false. The boundary must sanitize all user-controlled Drive result strings, not only content.

PR #52, opened 2026-08-26
feat(agent-runs): add durable checkpoint spine

FisherXZ:
## Summary

- add a canonical durable Agent Run and ordered semantic checkpoint store without repurposing scheduler receipts
- route chat, queued chat, and scheduled prompts through the same run identity while aggregating safe skill, source, artifact, usage, approval, and terminal metadata
- add versioned secret-scrubbing migrations, immutable goal fingerprints, restart interruption semantics, and the dependency-ordered Priority 0 implementation plan

## Test Plan

- [x] `make test` — 373 Python passed, 1 skipped; 336 GUI passed
- [x] `make build`
- [x] independent spec-compliance review
- [x] independent code-quality review with no Critical or Important findings

## Scope

This PR completes Priority 0 Slice A. Durable leases/restart resume, `ask_user`, bounded context, and terminal-delivery reserve remain in the next slices.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -37,6 +37,32 @@ The director's explicit instruction to send a reviewed Gmail message from Source
 **Outreach Outcome**
 What happened after outreach: no response, replied, interested, not a fit, needs follow-up, met, booked, or another recorded result. Outcomes update the person file and may move its sequence.
 
+## Agent work
+
+**Agent Job**
+A concrete outcome the director asks Sourcecado to produce. A job begins as instructions and does not require a skill.
+_Avoid_: Prompt, task, skill run
+
+**Agent Run**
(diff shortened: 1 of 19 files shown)


FisherXZ (commented):
## Verdict
**Ship.**

Context: This is Priority 0 Slice A. It adds a durable Agent Run and ordered semantic checkpoints around the existing turn loop, without turning scheduler receipts into the source of truth.

Why Ship: Chat, queued chat, and schedule now share one run identity. Secrets stay out of the durable record. Restart marks orphaned running work interrupted and leaves parked waits parked. Cancel still emits the same turn_stopped event as before.

## Blocking
None.

The failure I was most afraid of was the stop path. `_stopped` now goes through `_terminal` instead of `finish_stopped`. `finish_stopped` is still a wrapper around `send_terminal` with the same payload, and the inbox tests that require `turn_stopped` still pass.

PR #53, opened 2026-08-27
feat(agent-runs): add durable ownership and restart resume

FisherXZ:
## Summary

- add versioned Agent Run leases and compare-and-swap continuation checkpoints across model, tool, approval, and terminal boundaries
- resume safe interrupted work under the same canonical run identity while fencing duplicate owners and routing unknown external outcomes to review
- recover resumable runs at app startup, repair linked schedule receipts, and attach canonical run IDs to person-file ledger events
- mark Priority 0 Slice B complete with full verification evidence

## Test plan

- `make test` — 616 Python tests passed, 1 skipped; 336 GUI tests passed
- `make build` — TypeScript and Vite production build passed
- `git diff --check`

## Stack

This PR is stacked on #52 (`codex/product-validation-sprint`) and should be reviewed/merged after it.


FisherXZ (commented):
## Verdict
**Fix, then ship.**

Context: This is Priority 0 Slice B. It adds versioned Agent Run leases, continuation checkpoints, and restart resume on top of #52. Chat, schedule, and approval are supposed to keep one run identity after a crash.

Why Fix, then ship: The continuation and lease design is the right seam. Competing owners and unknown external outcomes are fenced in the tests I ran. A crash-restart of the live sidecar does not resume in-flight work. Store open keeps the dead owner's unexpired lease, and startup recovery never reaps it. The ticket's restart criterion only holds if the lease is already expired before the new process starts.

## Blocking
1. Crash-restart leaves a mid-model run `running` under the dead owner. I seeded a 30-second lease, opened the app the way startup does, and recovery launched 0 tasks. The provider was not called. After a 0.2-second lease, waiting in the running app still did not resume. Expiring the lease first, then opening the app, did resume the same run. Slice B says: on restart, resume safe incomplete work under the same run_id.

FisherXZ (commented):
Thanks. Reloaded Slice B and the startup-lease-reclaim plan at a835d4f. Seeded a 30-second `model_in_flight` lease under `crashed-owner`, opened a second store (the unexpired lease stayed), then opened the app the way startup does. Recovery launched 1 and the same run completed. Store-open still does not steal a live lease.

Issue #59, opened 2026-08-27
P0: Make DeepSeek tool-calling continuity and usage explicit


FisherXZ:
## What to build

Mode: **AFK**.

Make the active DeepSeek provider satisfy Sourcecado's complete model/tool contract. Thinking-mode responses must preserve provider-required continuity across tool-result substeps, emit typed answer and tool events, and finish with bounded usage metadata. Raw provider reasoning is transient protocol state, not a business artifact or a required UI surface.

This is provider correctness work. It must not make live chain-of-thought display a prerequisite for sourcing, approvals, restart recovery, or transcript replay.

## Acceptance criteria

- [ ] DeepSeek thinking mode and reasoning effort are configured explicitly.
- [ ] Provider chunks distinguish answer text, provider-required transient reasoning state, tool-call deltas, terminal metadata, and usage.
- [ ] Multi-step tool calling returns the exact provider-required continuity fields after each tool result.
- [ ] Streamed tool arguments are assembled and validated without losing call identity.
- [ ] Cancellation and provider failure close the current model step truthfully.
- [ ] Raw reasoning never enters durable conversation replay, Agent Run checkpoints, person files, memory, logs, approvals, or diagnostics.
- [ ] Reconnect or restart reports unavailable transient reasoning without fabricating recovered content.
- [ ] Contract tests cover text-only, one tool, multiple tools, multiple tool rounds, cancellation, malformed arguments, truncated streams, and provider errors.
- [ ] Terminal usage can feed the unified run ledger without exposing response content.
- [ ] Other provider clients can implement the same typed contract without DeepSeek-specific fields leaking into their interfaces.

## Blocked by

None — coordinate with the separate Agent Run port/reconciliation ticket before landing.

FisherXZ:
Status update — 2026-08-27\n\nPR #86 is open with CI green. Review found one blocking continuity defect: after a workspace tool such as fs_read or shell_exec, sanitization changes the durable tool payload shape, so the next user turn can miss the in-memory reasoning key and fail with unavailable continuation.\n\nThe assigned engineers are fixing that defect. This issue remains open in Sprint 1 until the corrected PR merges and verification is green.

FisherXZ:
[...]

PR #86, opened 2026-08-27
fix(desktop): preserve DeepSeek tool reasoning continuity

FisherXZ:
## Summary by CodeRabbit

* **New Features**
  * Added support for streamed reasoning, tool calls, completion status, and token usage.
  * Added DeepSeek provider support, including reasoning continuity and context handling.
  * Improved tool-call assembly and validation during streaming.
* **Bug Fixes**
  * Improved handling of malformed, incomplete, cancelled, and failed provider streams.
  * Preserved reasoning across tool interactions without including it in saved conversation content.

### DIFF
--- a/desktop/coworker/provider.py
+++ b/desktop/coworker/provider.py
@@ -4,7 +4,10 @@ from __future__ import annotations
 
 import json
 import os
+from collections import OrderedDict
+from copy import deepcopy
 from dataclasses import dataclass, field
+from hashlib import sha256
 from typing import Any, AsyncIterator, Optional, Protocol
 
 
@@ -15,10 +18,32 @@ class ToolCall:
     arguments: dict[str, Any] = field(default_factory=dict)
 
 
+@dataclass(frozen=True)
+class ToolCallDelta:
+    index: int
+    id: str | None = None
+    name_delta: str | None = None
+    arguments_delta: str | None = None
+
+
+@dataclass(frozen=True)
+class ModelUsage:
+    input_tokens: int
+    output_tokens: int
+    total_tokens: int
+    cached_input_tokens: int
+    uncached_input_tokens: int
+    reasoning_tokens: int
+
+
 @dataclass
 class StreamChunk:
     text_delta: str | None = None
+    transient_reasoning_delta: str | None = None
+    tool_call_deltas: list[ToolCallDelta] | None = None
     tool_calls: list[ToolCall] | None = None
+    finish_reason: str | None = None
+    usage: ModelUsage | None = None
 
 
 class StreamProvider(Protocol):
@@ -29,6 +54,7 @@ class StreamProvider(Protocol):
         *,
         messages: list[dict[str, Any]],
         tools: list[dict[str, Any]] | None = None,
+        context_id: str | None = None,
     ) -> AsyncIterator[StreamChunk]: ...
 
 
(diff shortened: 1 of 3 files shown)


FisherXZ (commented):
## Verdict
**Fix, then ship.**

Context: This is issue #59. DeepSeek thinking mode now keeps tool-call continuity in process memory and does not write chain-of-thought into the transcript.

Why Fix, then ship: Same-turn tool loops work, and a follow-up after `now()` works. A follow-up after `fs_read` or `shell_exec` does not. Workspace sanitization rewrites the stored tool payload, the continuity hash misses, and the next user message errors even though the reasoning is still in memory. #59 says not to persist raw reasoning. It does not ask to fail a live chat after a redacted workspace tool.

## Blocking
1. After `fs_read` or `shell_exec`, the next user message in the same session fails with `deepseek transient reasoning is unavailable for continuation` on the same provider instance. Durable tool payloads lose file content or command output. The continuity key hashes those payloads, so lookup misses while the cache still holds the reasoning. I ran `run_turn` twice against the same `DeepSeekProvider`: `now()` returned ok on the second turn; `fs_read` and `shell_exec` returned status error and wrote that message on the error event. Hash fields that survive workspace sanitization, or key the cache from the persisted message shape too. Do not persist `reasoning_content`.

FisherXZ (commented):
## Verdict
**Ship.**

Context: Follow-up to the workspace-sanitization blocker on #59. The continuity key now ignores stored tool payloads that workspace persist rewrites.

Why Ship: I reran the same two-turn `run_turn` cases. `now()`, `fs_read`, and `shell_exec` all returned ok on the follow-up turn and replayed reasoning. Durable transcripts still had no `reasoning_content`. A fresh provider still failed closed before a network request. Python sidecar tests passed locally, including the two new follow-up tests.

## Blocking
None. The failure I was most afraid of was the next user message after `fs_read` still erroring. I ran that path again after the fix. Status was ok and the follow-up request carried the original in-memory reasoning.

FisherXZ (commented):
## Verdict
**Ship.**

## Summary
DeepSeek thinking mode keeps tool-call continuity in process memory. Same-turn tool loops and a follow-up after `now()`, `fs_read`, and `shell_exec` replay in-memory reasoning. Durable transcripts still have no `reasoning_content`. A fresh provider still fails closed before a network request.

The first head hashed stored tool payloads. Workspace persist rewrites `fs_read` content and `shell_exec` arguments, so a follow-up after those tools missed the cache. The merged head ignores those rewritten fields.

## Map

```mermaid
flowchart LR
  A[User turn] --> B[DeepSeek astream]
  B --> C[Assemble tool calls]
  C --> D[Execute tools]
  D --> E[Workspace persist]
  E --> F[Stored payload rewritten]
  F --> G[Next user turn]
  G --> H[Continuity key]
  H --> I[In-memory reasoning cache]
  F -. key ignores rewritten payloads .-> H
  I -->|miss on a fresh process| J[Fail closed before network]
  I -->|hit on the same process| K[Replay reasoning_content on the wire]
```

Why Ship: #59 requires continuity across tool-result substeps and forbids raw reasoning in durable replay. I reran `tests/test_provider.py` on the merged head: 19 passed, including the two-turn `fs_read` and `shell_exec` follow-ups. Those tests assert the follow-up request carries the original in-memory reasoning and that `reasoning_content` is absent from the durable transcript.

## Blocking
None.

The failure I was most afraid of was the next user message after `fs_read` still erroring. That path is covered on this head and it passed.

PR #85, opened 2026-08-27
feat(desktop): add typed runtime telemetry foundation

FisherXZ:
## Summary\n\n- add closed, versioned operational schemas for provider, agent-turn, tool, retry, compaction, usage, cost, and terminal telemetry\n- add inert no-op and deterministic in-memory adapters with passive failure containment\n- add single-settlement spans with parentage, concurrent ordering, and inert post-settlement behavior\n- add content and secret exclusion tests so prompts, responses, arguments, results, raw errors, credentials, and reasoning cannot enter telemetry\n- add current-run token, context, elapsed-time, retry, compaction, and estimated-cost projection\n\n## Verification\n\n- focused telemetry suite: 47 passed\n- make test-python: 511 passed, 2 skipped\n- pre-existing Starlette/httpx deprecation warning remains\n\n## Review focus\n\n- schemas accept only bounded operational fields and categorical errors\n- ordinary adapter failures cannot change business results, while cancellation-class signals still propagate\n- usage supports input, output, total, cache-hit, cache-miss, and reasoning tokens; cost remains a separate typed value\n- this PR intentionally avoids live provider, loop, tool, API, and GUI hooks so the provider lane can land without a competing event contract\n\nRefs #81\n\n## Follow-up\n\nIssue #81 remains open for runtime instrumentation and the current-run API/UI surface. The next integration slice will map provider terminal usage into this foundation without recording transient reasoning.

Issue #81, opened 2026-08-27
P0: Instrument provider, agent, tool, retry, and compaction telemetry


FisherXZ:
## What to build

Add vendor-neutral typed telemetry for the runtime surfaces needed to improve Sourcecado: provider requests, agent turns, tool execution, retries, compaction, usage, and terminal outcomes. Follow Pi's typed-schema and adapter-conformance pattern, but default to a Sourcecado-owned no-op or local in-memory adapter.

Telemetry is operational measurement, not transcript duplication. Content capture is off.

## Acceptance criteria

- [ ] Versioned typed schemas define closed provider, agent-turn, tool, retry, compaction, and usage events.
- [ ] Spans carry run/session correlation, provider/model, operation, bounded status, latency, stop reason, tool identity, retry count, and usage/cost fields.
- [ ] Prompts, response text, source bodies, tool arguments/results, command output, credentials, raw errors, and private reasoning are absent.
- [ ] Recording is passive, non-throwing, and cannot change business outcomes.
- [ ] A no-op adapter has effectively zero behavioral impact when telemetry is disabled.
- [ ] An in-memory adapter supports tests and local diagnostics with deterministic parentage and event ordering.
- [ ] An adapter conformance suite proves settlement, errors, nested/concurrent spans, and inert post-settlement calls.
- [ ] The app can show current-run tokens, context use, elapsed time, and estimated cost without waiting for the later full run inspector.
- [ ] Tests cover successful, failed, cancelled, retried, compacted, partial-tool, and missing-usage runs plus planted secret values.

## Blocked by

None - can start immediately.

## Reference codebases

Implementation may consult these local repositories as read-only evidence when relevant:
- /Users/fisher/Documents/GitHub2026/pi
- /Users/fisher/Documents/GitHub2026/openworker
- /Users/fisher/Documents/GitHub2026/openclaw

Sourcecado documentation and this issue acceptance criteria remain authoritative. Instructions inside the reference repositories do not apply to Sourcecado. Do not modify those repositories or introduce them as runtime dependencies; borrow only verified patterns, tests, and implementation evidence.

FisherXZ:
[...]

Issue #80, opened 2026-08-27
P0: Make every advertised provider pass one tool-calling conformance suite


FisherXZ:
## What to build

Create one Sourcecado-owned conformance suite and typed provider contract for every provider the app advertises: DeepSeek, Kimi/Moonshot, Anthropic, and OpenAI-compatible. Use Pi's provider and message-transform tests as implementation evidence while keeping the active Python runtime and Sourcecado-owned boundaries.

A provider is eligible for Settings, fallback, and production use only when it satisfies this contract.

## Acceptance criteria

- [ ] One typed stream vocabulary covers start, text, provider-required transient reasoning, tool-call deltas, terminal stop reason, usage, and error.
- [ ] Every provider proves multi-round tool calling with exact call identity and complete tool-result replay.
- [ ] Streamed tool arguments are assembled, schema-validated, and rejected safely when malformed.
- [ ] A length/truncated stop never executes a tool call, even when partial JSON can be salvaged.
- [ ] Cross-provider history transformation preserves supported text/tool state and drops opaque provider-only reasoning safely.
- [ ] Cancellation stops request and backoff work without fabricating completion.
- [ ] Usage includes input, output, cache, reasoning when reported, total tokens, latency, and estimated cost when pricing is known.
- [ ] Unsupported or partially configured providers fail a verification check and are not eligible fallback targets.
- [ ] Deterministic fake-provider fixtures cover the full matrix; opt-in live smoke records provider, model, prompt version, and nondeterminism.
- [ ] Settings reports verified capabilities and setup failures without exposing credentials.

## Blocked by

- #59

## Reference codebases

Implementation may consult these local repositories as read-only evidence when relevant:
- /Users/fisher/Documents/GitHub2026/pi
- /Users/fisher/Documents/GitHub2026/openworker
- /Users/fisher/Documents/GitHub2026/openclaw

Sourcecado documentation and this issue acceptance criteria remain authoritative. Instructions inside the reference repositories do not apply to Sourcecado. Do not modify those repositories or introduce them as runtime dependencies; borrow only verified patterns, tests, and implementation evidence.

FisherXZ:
[...]

PR #87, opened 2026-08-27
feat(desktop): wire live runtime telemetry

FisherXZ:
## Summary\n\n- create one in-memory telemetry recorder at the sidecar composition root\n- instrument authoritative agent-turn, provider-request, tool-execution, approval, cancellation, partial, step-limit, and manual-retry paths with session/run parentage\n- map provider usage and terminal state into the closed content-free telemetry vocabulary from PR #85\n- add an authenticated current-run telemetry endpoint\n- show a compact current-run metrics strip in chat for tokens, context use, elapsed time, retries, compactions, and cost availability\n- rebuild explicit API and GUI allowlists so prompts, responses, reasoning, arguments, results, command output, credentials, and raw errors cannot escape\n\n## Verification\n\n- focused Python runtime, approval, and retry suite: 108 passed\n- focused GUI suite: 65 passed\n- make test: 546 Python passed, 2 skipped; 363 GUI passed\n- make build: passed\n- existing Starlette/httpx deprecation warning and Vite large-chunk advisory remain\n\n## Review focus\n\n- telemetry cannot change business outcomes or leak content\n- all run and tool paths settle exactly once with authoritative session/run identity\n- metrics remain isolated when navigating between threads\n- manual retry increments only when execution actually occurs\n- cost and context-window percentages remain explicitly unavailable until provider metadata supplies them\n\nRefs #81\n\n## Follow-up\n\nIssue #81 remains open until the #80 provider contract supplies stable provider identity, context-window metadata, and estimated-cost data where pricing is known.

### DIFF
--- a/desktop/coworker/telemetry/adapters.py
+++ b/desktop/coworker/telemetry/adapters.py
@@ -74,6 +74,7 @@ class InMemoryTelemetryAdapter:
             for record in self.records
             if isinstance(record, SpanStartedRecord)
             and isinstance(record.span, AgentTurnSpan)
+            and record.span.operation == "agent.turn"
             and record.context.session_id == session_id
         ]
         if not starts:
--- a/desktop/coworker/telemetry/adapters.py
+++ b/desktop/coworker/telemetry/adapters.py
@@ -8,6 +8,7 @@ from threading import Lock, RLock
 from typing import Protocol
 
 from coworker.telemetry.schema import (
+    AgentTurnSpan,
     CostEstimate,
     ErrorKind,
     SpanEventRecord,
@@ -66,6 +67,22 @@ class InMemoryTelemetryAdapter:
         from coworker.telemetry.metrics import current_run_metrics
 
         return current_run_metrics(self.records, run_id=run_id, now_ns=now_ns)
+
+    def current_session_metrics(self, *, session_id: str, now_ns: int):
+        starts = [
+            record
+            for record in self.records
+            if isinstance(record, SpanStartedRecord)
+            and isinstance(record.span, AgentTurnSpan)
+            and record.context.session_id == session_id
+        ]
+        if not starts:
+            return None
+        latest = max(starts, key=lambda record: record.sequence)
+        return self.current_run_metrics(
+            run_id=latest.context.run_id,
+            now_ns=now_ns,
+        )
 
 
 class TelemetryRecorder:
--- a/desktop/coworker/turn.py
+++ b/desktop/coworker/turn.py
@@ -63,6 +63,13 @@ def _telemetry_provider_span(provider: Any) -> ProviderSpan:
         )
 
 
+def _telemetry_tool_span(name: str) -> ToolSpan:
+    try:
+        return ToolSpan(tool_name=name, operation="tool.execute")
+    except ValueError:
+        return ToolSpan(tool_name="unknown", operation="tool.execute")
+
+
 def _telemetry_usage(usage: ModelUsage) -> UsageEvent:
     return UsageEvent(
         input_tokens=usage.input_tokens,
@@ -934,9 +941,7 @@ async def run_turn(
                         "started_at": _now_iso(),
                     }
                 )
-                tool_span = turn_span.child(
-                    ToolSpan(tool_name=call.name, operation="tool.execute")
-                )
+                tool_span = turn_span.child(_telemetry_tool_span(call.name))
                 try:
                     kw = {
                         k: v for k, v in execute_kwargs.items() if not k.startswith("_")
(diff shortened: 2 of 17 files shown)


desktop/coworker/turn.py
```diff
@@ -834,6 +934,9 @@ async def _stopped(
                         "started_at": _now_iso(),
                     }
                 )
+                tool_span = turn_span.child(
+                    ToolSpan(tool_name=call.name, operation="tool.execute")
+                )
```
FisherXZ: `ToolSpan(...)` is evaluated before `child()`, so a closed-schema miss fails the turn even when the recorder is no-op. Approval and recovery already fall back to `unknown`.

```suggestion
                try:
                    tool_schema = ToolSpan(
                        tool_name=call.name, operation="tool.execute"
                    )
                except ValueError:
                    tool_schema = ToolSpan(
                        tool_name="unknown", operation="tool.execute"
                    )
                tool_span = turn_span.child(tool_schema)
```

desktop/coworker/telemetry/adapters.py
```diff
@@ -67,6 +68,22 @@ def current_run_metrics(self, *, run_id: str, now_ns: int):
...
         return current_run_metrics(self.records, run_id=run_id, now_ns=now_ns)
 
+    def current_session_metrics(self, *, session_id: str, now_ns: int):
+        starts = [
+            record
+            for record in self.records
+            if isinstance(record, SpanStartedRecord)
+            and isinstance(record.span, AgentTurnSpan)
+            and record.context.session_id == session_id
+        ]
```
FisherXZ, line 79: Latest `AgentTurnSpan` includes `agent.approval` and `agent.recovery`, so an older inbox Allow replaces the current chat run in the header strip.

```suggestion
        starts = [
            record
            for record in self.records
            if isinstance(record, SpanStartedRecord)
            and isinstance(record.span, AgentTurnSpan)
            and record.span.operation == "agent.turn"
            and record.context.session_id == session_id
        ]
```

FisherXZ (commented):
## Verdict
**Fix, then ship.**

## Summary
This wires the PR #85 telemetry foundation into the live sidecar. One in-memory recorder sits at `create_app`. Agent turns, provider requests, tool executions, approvals, cancellations, partials, step-limit stops, and manual retries now settle with session/run identity. Chat polls an authenticated current-run endpoint and renders a compact metrics strip. API and GUI allowlists keep prompts, responses, arguments, results, and raw errors out.

The recorder seam belongs here. Two live paths still break the #81 rules this slice is supposed to prove.

## Map
```mermaid
flowchart TD
  user[User turn] --> runTurn[run_turn]
  runTurn --> turnSpan[AgentTurnSpan agent.turn]
  turnSpan --> providerSpan[ProviderSpan]
  turnSpan --> toolExec[Construct ToolSpan then execute]
  toolExec -.->|allowed name fails symbol regex| crash[Turn status error]
  runTurn --> settle[Settle spans once]
  settle --> api["GET telemetry/current"]
  api --> latest[current_session_metrics latest AgentTurnSpan]
  latest --> strip[Chat metrics strip]
  allow[Later inbox Allow] --> approval[AgentTurnSpan agent.approval]
  approval -.->|newer sequence older run_id| latest
```

Why Fix, then ship: #81 requires recording to be passive and current-run metrics to be the current run. The composition root and content allowlists hold. I reran the worktree suites: 546 Python passed, 2 skipped; 363 GUI passed. A permitted MCP tool name with a space still fails the turn. A later Allow on an older inbox item still replaces the current-run strip.

## Blocking
1. A live tool name that `decide()` allows, but the closed telemetry symbol regex rejects, fails the whole turn as `error` instead of a tool miss. `mcp__granola__list meetings` returned `{'status': 'error', 'text': ''}` with the in-memory recorder and with the default no-op recorder. The same name without telemetry construction would continue as `partial`. `ToolSpan` is built in the business path before `child()`, so even a disabled recorder still throws. Approval and recovery already catch this `ValueError` and fall back to `unknown`.

2. After a later chat turn, Allow on a leftover inbox item from an earlier run makes `GET /v1/sessions/{sid}/telemetry/current` report that earlier run. I ran turn 1, then turn 2 (`run_cb2f6b1caf3546029d59dd65a933e325`), then Allow on the parked turn-1 item. The current-run payload came back as `run_a27c33d0a719435ab4c0991e0c259614`. `current_session_metrics` picks the latest `AgentTurnSpan` of any operation, so `agent.approval` and `agent.recovery` can replace the current chat run.

FisherXZ (commented):
Round 2 on add9e9a. Loaded #81, the merge-base diff against origin/main, and the two follow-up commits. A live `mcp__granola__list meetings` turn still finishes partial with recovered text on the in-memory recorder and the default no-op recorder. After a later chat turn, Allow on the leftover first-run inbox item left `GET /v1/sessions/{sid}/telemetry/current` on the second run. Worktree suites: 548 Python passed, 2 skipped; 363 GUI passed.

Issue #71, opened 2026-08-27
P0: Retry and fail over model requests without replaying completed effects


FisherXZ:
## What to build

Add bounded, classified model-request retry and provider failover on top of the typed provider contract. Retry only the current model request or safe continuation step. Never replay a completed tool, approval, enrichment, send, calendar write, person mutation, filesystem write, or shell action.

Record every attempt and usage result against the canonical Agent Run so the operator can distinguish transient recovery from duplicated work.

## Acceptance criteria

- [ ] Retryable provider failures include bounded transport errors, timeouts, 408, 409, 429, and selected 5xx responses.
- [ ] Retry honors bounded Retry-After values and otherwise uses capped exponential backoff with jitter.
- [ ] Authentication, validation, policy, malformed tool, and non-retryable provider errors fail immediately.
- [ ] Failover occurs only after the active provider's bounded attempts are exhausted and preserves the same semantic run/checkpoint identity.
- [ ] A retry or failover after completed tools continues from their durable results and never executes them again.
- [ ] Ambiguous partial streams are reconciled or surfaced for review rather than concatenated with a fresh answer.
- [ ] Attempt count, provider/model, latency, input/output/cache usage, reason category, and final selection enter the run ledger.
- [ ] The UI reports recovery or final provider failure without exposing raw provider bodies.
- [ ] Tests cover rate limits, Retry-After, timeout, truncated stream, failover, cancellation during backoff, tool-before-failure, write-before-failure, and exhausted chains.

## Blocked by

- #59
- #80
## Sprint 1 provider eligibility contract

Authoritative plan: [Sourcecado Sourcing OS Release Program — Execution Plan](https://app.notion.com/p/3c9d64acf06681af91dbee2d65715f70)

A provider may enter a failover chain only after passing the complete Sourcecado provider conformance suite for answer text, multi-round tools, exact call identity, malformed arguments, cancellation, terminal metadata, usage, context budget, and provider-required continuation fields.

### Additional release acceptance criteria

- [ ] An adapter with missing or incompatible tool behavior is rejected from failover configuration before a run starts.
- [ ] Failover eligibility and rejection reason are inspectable without exposing provider payloads.
- [ ] Cross-provider tests prove completed tool results remain authoritative and are not re-executed.

## Reference codebases

Implementation may consult these local repositories as read-only evidence when relevant:
- /Users/fisher/Documents/GitHub2026/pi
- /Users/fisher/Documents/GitHub2026/openworker
[...]

PR #91, opened 2026-08-27
feat(desktop): add bounded provider retry and failover

FisherXZ:
## Summary\n\n- classify retryable transport, timeout, 408, 409, 429, and selected 5xx provider failures\n- honor bounded numeric and HTTP-date Retry-After with capped exponential jitter\n- fail immediately on authentication, validation, policy, protocol, and malformed-response errors\n- fail over only after current-provider attempts are exhausted and only to verified eligible providers\n- continue from durable completed tool results without replaying tools, approvals, sends, filesystem writes, shell actions, or other effects\n- surface ambiguous partial streams for review instead of concatenating fresh output\n- record typed attempt, provider/model, latency, usage/cache, reason, and final selection telemetry\n- add content-free provider recovery and final-failure UI with cancellation-aware request/backoff behavior\n- preserve Kimi plain-history and fallback tool-call continuity without durable reasoning\n\n## Verification\n\n- focused Python: 234 passed\n- focused GUI: 82 passed\n- make test: 668 Python passed, 3 skipped; 382 GUI passed\n- make build: passed\n- independent post-rebase review: SHIP; 191 focused Python and 58 GUI tests passed\n- git diff --check: passed\n\nCloses #71\n\n## Residual risk\n\nCredentialed live-provider smoke was not run. Retry schedules remain process-local, while completed tool results remain durable and authoritative after restart.

### DIFF
--- a/desktop/coworker/turn.py
+++ b/desktop/coworker/turn.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import random
 import threading
 import uuid
 from datetime import UTC, datetime
@@ -22,12 +23,23 @@ from coworker.provider import (
     ToolCall,
     provider_model_metadata,
 )
+from coworker.provider_retry import (
+    ProviderRequestCancelled,
+    RecoveryAction,
+    RetryController,
+    RetryPolicy,
+    cancellable_stream,
+    compatible_failover_chain,
+    safe_provider_failure_message,
+)
 from coworker.store import ConversationStore
 from coworker.telemetry import (
     AgentTurnSpan,
     CostEstimate,
     ErrorKind,
     ProviderSpan,
+    RetryEvent,
+    RetryReason,
     StopReason,
     TelemetryRecorder,
     TraceContext,
@@ -613,6 +625,10 @@ async def run_turn(
     identity: TurnIdentity | None = None,
     control: RunControl | None = None,
     telemetry: TelemetryRecorder | None = None,
+    failover_providers: tuple[Any, ...] = (),
+    retry_policy: RetryPolicy | None = None,
+    retry_sleep: Callable[[float], Awaitable[None]] | None = None,
+    retry_random: Callable[[], float] = random.random,
 ) -> dict[str, Any]:
     workspace_runtime = execute_kwargs.get("workspace_runtime")
     events = TurnEventStream(
@@ -725,8 +741,18 @@ async def run_turn(
         )
         return {"status": "error", "text": ""}
 
+    provider_chain = (provider,) + tuple(
+        candidate
+        for candidate in failover_providers
+        if candidate is not provider
+    )
+    selected_provider = provider
+    retry_count = 0
+
     last_text = ""
     had_tool_failure = False
+    turn_error_kind = ErrorKind.INTERNAL
+    turn_error_message: str | None = None
     history: list[dict[str, Any]] = []
     try:
         raw = store.load(sid)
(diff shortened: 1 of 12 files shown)


FisherXZ (commented):
## Verdict
**Fix, then ship.**

## Summary
The sidecar classifies retryable provider failures, honors bounded Retry-After including HTTP-date, fails closed on auth, validation, and protocol errors, and fails over only after the active provider is exhausted. Completed tools, approvals, and sends stay durable and are not replayed. Visible truncated streams go to review instead of concatenation. Recovery events stay content-free.

The miss is what counts as a partial stream. Any non-START chunk, including hidden DeepSeek or Kimi thinking, blocks retry and failover.

## Map
```mermaid
flowchart TD
  req[Model request] --> start[START chunk]
  start --> think[Hidden reasoning delta]
  think --> gate{Meaningful stream?}
  gate -->|now: any non-START chunk| review[REVIEW]
  review --> fail[Turn fails]
  fail --> unused[Fallback unused]
  gate -.->|needed: text or tool payload only| retry[Retry or failover]
  retry --> ok[Continue with no visible concatenation]
```

Why Fix, then ship: The retry and failover design belongs in this codebase. #71 still requires timeout and 429 recovery when no user-visible answer exists. I reproduced START plus reasoning plus 429 on this head: one attempt, no failover. Suites held: 668 Python passed, 3 skipped; 382 GUI passed.

## Blocking
1. A DeepSeek or Kimi request that times out or hits 429 while streaming hidden thinking never retries and never fails over. The operator is told to review a partial response that was never shown. #71 requires bounded retry of timeouts and 429, and review only when concatenating a fresh visible answer would be ambiguous.

desktop/coworker/turn.py
```diff
@@ -755,97 +781,202 @@ async def _stopped(
...
+                    async for chunk in cancellable_stream(
+                        attempt_provider.astream(**stream_kwargs),
+                        cancel_event=(
+                            control.cancel_requested if control is not None else None
+                        ),
+                    ):
+                        if chunk.kind is StreamKind.START and chunk.start is not None:
+                            _ensure_provider_span(chunk.start)
+                            continue
+                        if chunk.kind is not None:
```
FisherXZ, line 839: Hidden thinking is not a user-visible answer. This treats every non-START chunk as a partial stream, so a DeepSeek timeout during `reasoning_content` skips retry and failover.

```suggestion
                        if chunk.kind in {StreamKind.TEXT, StreamKind.TOOL_DELTA, StreamKind.TOOL_CALLS}:
```

FisherXZ (commented):
## Verdict
**Ship.**

## Summary
Round 1 blocked hidden DeepSeek and Kimi thinking as a partial stream. Head 41fad66 skips `StreamKind.REASONING` before the meaningful-stream gate. Timeout and 429 during thinking retry and fail over. Visible text or tool payload still goes to review.

## Map
```mermaid
flowchart TD
  req[Model request] --> start[START chunk]
  start --> think[Hidden reasoning delta]
  think --> gate{Meaningful stream?}
  gate -->|text or tool payload| review[REVIEW]
  gate -->|reasoning only| retry[Retry or failover]
  retry --> ok[Continue with no visible concatenation]
  review --> fail[Turn fails]
```

Why Ship: #71 requires bounded retry of timeouts and 429 when no user-visible answer exists. I reran START plus hidden reasoning plus 429/timeout on this head. It failed over. Suites held: 670 Python passed, 3 skipped; 382 GUI passed.

## Blocking
None.

The failure I was most afraid of was START plus reasoning plus 429 still skipping failover. That path is covered on this head and it passed.

Issue #55, opened 2026-08-27
P0: Derive prompt capabilities from the effective runtime tool catalog


FisherXZ:
## What to build

Mode: **AFK** after the prompt's capability section shape is approved in #54.

Make the effective runtime tool catalog and permission policy the sole source of truth for model capabilities. Remove the hand-maintained tool inventory from the system kernel. Persona tool declarations must either resolve to real registered tools or fail validation; ignored or nonexistent declarations must not survive as apparent configuration.

Reuse the contract principle from #17 and #26: prompt capability claims, registered tools, and granted permission classes must be derived from the same validated inventory. Implement it in the active Python/FastAPI desktop runtime, not the archived hosted stack.

## Acceptance criteria

- [ ] The active system prompt no longer contains a manually maintained list of every tool.
- [ ] Capability guidance is derived from the exact tool schemas and effective permission decisions supplied to the current run.
- [ ] Persona-declared tool names are either removed as redundant or validated against the runtime registry.
- [ ] An unknown or misspelled persona tool fails clearly during startup or persona loading.
- [ ] Sourcing and buddy personas cannot advertise capabilities they do not receive.
- [ ] Changes to a tool name, availability, or approval class fail a contract test until prompt-facing metadata is reconciled.
- [ ] Diagnostics can report the effective model-facing tool names without exposing secrets or raw schemas.

## Blocked by

- #54

## Prior work to reuse

- #17 — prompt commitments and allowed permission classes were wired together.
- #26 — persona tool declarations were validated against the implemented inventory.

## Sprint 2 execution contract

Authoritative plan: [Sourcecado Sourcing OS Release Program — Execution Plan](https://app.notion.com/p/3c9d64acf06681af91dbee2d65715f70)

The effective tool surface is a per-run allowlist. The registry is the set of implementable tools, not permission to give every run every tool. Persona declarations may narrow the registry but may not broaden it.

### Additional release acceptance criteria

- [ ] The sourcing and buddy personas receive explicit effective tool subsets derived from registered tools and current permission policy.
- [ ] Filesystem write and shell capabilities are absent unless the run has the corresponding Sourcecado-owned workspace contract and operator grant.
- [ ] A persona cannot gain a tool by naming it in frontmatter or prompt prose.
- [ ] Diagnostics report the effective tool names and approval classes without exposing schemas, arguments, or secrets.

## Reference codebases

Implementation may consult these local repositories as read-only evidence when relevant:
[...]

PR #97, opened 2026-08-27
feat(desktop): derive effective tools from runtime policy

FisherXZ:
## Summary\n\n- build a fresh effective tool catalog for every scheduled and interactive run\n- derive model-facing tools from registered schemas, persona policy, connector availability, workspace grants, and permission classes\n- attach auto, approval_required, or conditional guidance to exact schemas\n- remove persona frontmatter tool declarations as an authority source\n- let custom declarations narrow but never broaden, failing clearly on unknown names\n- enforce a runtime-owned narrower buddy subset with registry-drift tests\n- require workspace authority before exposing filesystem or shell tools\n- reject and persist out-of-catalog provider calls without execution\n- remove the cached duplicate app.state.openai_tools source\n- expose only effective tool names and approval classes in diagnostics\n\n## Verification\n\n- focused effective-tool integration: 128 passed\n- catalog and permission contract: 25 passed\n- make test: 746 Python passed, 3 skipped; 382 GUI passed\n- make build: passed\n- git diff --check: passed\n- rebased onto current main after approved prompt activation\n\nCloses #55\n\n## Residual consideration\n\nConnector and workspace availability is sampled at turn start and applies to the next run. Conditional guidance remains schema-level; concrete arguments and grants determine final runtime permission.

### DIFF
--- a/desktop/coworker/effective_tools.py
+++ b/desktop/coworker/effective_tools.py
@@ -1,0 +1,192 @@
+"""Validated per-run model-facing tool catalogs."""
+
+from __future__ import annotations
+
+from copy import deepcopy
+from dataclasses import dataclass
+from typing import Any, Iterable
+
+from coworker.persona import ManifestError, Persona
+from coworker.permissions import model_approval_class
+from coworker.workspace_runtime import WORKSPACE_TOOL_NAMES, WorkspaceRuntime
+
+
+class ToolCatalogError(ValueError):
+    pass
+
+
+@dataclass(frozen=True, slots=True)
+class ToolAvailability:
+    gmail: bool = True
+    drive: bool = True
+    calendar: bool = True
+    apollo: bool = True
+    web: bool = True
+
+
+@dataclass(frozen=True, slots=True)
+class ToolCapability:
+    name: str
+    approval_class: str
+
+
+@dataclass(frozen=True, slots=True)
+class EffectiveToolCatalog:
+    schemas: tuple[dict[str, Any], ...]
+    capabilities: tuple[ToolCapability, ...]
+
+    @property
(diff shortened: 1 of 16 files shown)


FisherXZ (commented):
## Verdict
**Ship.**

## Summary
Each scheduled and interactive turn now builds a fresh model-facing catalog from registered schemas, runtime persona policy, optional declaration narrowing, connector availability, workspace grants, and schema-level approval classes. Built-in personas no longer declare tools. Buddy is a runtime-owned subset that cannot be broadened by frontmatter or prose. Unknown declarations fail at compose or startup. Provider calls outside that catalog are persisted as failures and never execute. Prompt diagnostics expose only names and approval classes.

## Map
```mermaid
flowchart TD
  turnStart[Turn start] --> compose[Compose effective catalog]
  compose --> toolsParam[Send annotated schemas to provider]
  toolsParam --> call{Provider tool call}
  call -->|name in catalog| decide[Runtime decide]
  decide --> exec[Execute or ask]
  call -->|name not in catalog| reject[Persist failed tool result]
  reject --> noExec[Never execute]
```

Why Ship: #55 required the per-run allowlist to be the sole model capability surface, with fail-closed unknown names and no filesystem or shell tools without a grant. I loaded issue #55, the sprint execution plan, and the diff against merge-base `418cd54`. I ran `desktop/.venv/bin/pytest -q` on this head: 746 passed, 3 skipped.

## Blocking
None.

The failure I was most afraid of was a disconnected Gmail still receiving `gmail_send`, or an out-of-catalog `remember` still writing memory. Composition without connectors excludes Gmail, Drive, Calendar, Apollo, web, filesystem, and shell. The empty-catalog `remember` path persisted a failure and left the memory store empty.

PR #102, opened 2026-08-28
fix(desktop): normalize legacy scheduled-run statuses truthfully

### LINKED ISSUE #40: P2: Normalize legacy scheduled-run statuses and receipt truthfulness
## What to build

Mode: **AFK**.

Restore truthful scheduled-run receipts for legacy rows. Production rows with status `ok` and complete results currently normalize to `failed`, so Chrome shows “Failed / 0ms” beside successful sourcing reviews. Normalize or migrate legacy statuses and make unknown statuses visibly indeterminate rather than falsely failed.

## Acceptance criteria

- [ ] Legacy `ok` maps to canonical `success`, either through migration or API normalization.
- [ ] A production-shaped restored `ok` row renders as Completed.
- [ ] Genuinely unknown status renders as Needs review/Unknown, not Failed.
- [ ] Receipt status agrees with any available scheduled transcript terminal event.
- [ ] Legacy zero-duration rows omit duration or display Legacy run rather than `0ms`.
- [ ] New scheduler writes use only canonical status vocabulary.
- [ ] Tests cover legacy `ok`, canonical statuses, unknown status, waiting approval, partial, and interrupted/crash recovery.

## Blocked by

None - can start immediately.

## Reference codebases

FisherXZ:
Closes #40.

## Domain words

- **Receipt** — the stored record of one scheduled run, shown on the Scheduled page.
- **Canonical status** — the six values the sidecar and the client agree on: `running`, `success`, `failed`, `waiting_approval`, `partial`, `interrupted`.
- **Legacy row** — a run row written before that shared vocabulary existed.
- **Reconciler** — the startup pass that closes runs a crashed process left marked `running`.

## Problem

Scheduled runs that succeeded were displayed as "Failed / 0ms".

## Before and after

| Stored row | Before | After |
|---|---|---|
| legacy `status = "ok"`, full result | Failed | Completed |
| unrecognized status value | Failed | Needs review |
| `duration_ms = 0` on a legacy row | `0ms` | `Legacy run` |
| crash after transcript wrote `turn_end`/`complete` | Interrupted | Completed |
| crash with no terminal transcript event | Interrupted | Interrupted (unchanged) |

## Cause

Two separate defects.

First, `desktop/surfaces/gui/src/api.ts` normalized run status with `SCHEDULE_RUN_STATUSES.has(rawStatus) ? rawStatus : "failed"`. The set does not contain `"ok"`. Legacy rows therefore became `failed`. The same line also turned any genuinely unknown value into `failed`, so two different situations produced one wrong answer.

Second, `_reconcile_orphaned_runs` in `desktop/coworker/store.py` stamped every row still marked `running` as `interrupted` on restart. A crash can land after the transcript records a terminal event but before the runs table is updated. In that window the outcome is already known, and the reconciler discarded it. This is beyond what the ticket described.

## Fix

The legacy alias is migrated once at store init, next to the existing runs-table backfill: `UPDATE runs SET status = 'success' WHERE status = 'ok'`. Migration rather than read-time mapping, because read-time mapping leaves the wrong value on disk for every future consumer to re-guess.

An unrecognized status now becomes a client-only `unknown`, rendered "Needs review". It is deliberately kept out of `SCHEDULE_RUN_STATUSES` on both sides. The sidecar never writes it, so it does not belong in the shared contract, and the cross-language vocabulary parity test still holds.

The reconciler now reads the session transcript for each orphaned row and uses its terminal event when there is one. With no terminal event it still writes `interrupted`.

## Measurements

Python suite:

```
732 passed, 3 skipped, 1 warning in 40.01s
```

GUI suite:

```
Test Files  43 passed (43)
     Tests  385 passed (385)
```

GUI build: exit 0, 992 modules transformed.

Both suites were re-run independently by the reviewing session on the pushed commit, not only by the implementing session.

## Acceptance criteria

| Criterion | Test |
|---|---|
| Legacy `ok` maps to `success` | `test_legacy_ok_run_normalizes_to_success_on_restart` |
| Restored `ok` row renders Completed | same, plus `scheduledPage.test.tsx` receipt-rendering test |
| Unknown status renders Needs review | `scheduledApi.test.ts` unknown-status test, `scheduledPage.test.tsx` needs-review test |
| Status agrees with transcript terminal event | `test_reopen_recovers_true_success_from_the_scheduled_transcript`, `test_reopen_recovers_true_failure_from_the_scheduled_transcript` |
| Zero-duration shows Legacy run | `scheduledPage.test.tsx` legacy-duration test |
| New writes use canonical vocabulary | `test_scheduled_run_statuses_stay_inside_the_shared_contract` (pre-existing, unchanged) |
| Coverage across all status kinds | the above, plus pre-existing waiting-approval and partial tests, unmodified |

## Not in this PR

- No change to `scheduler.py`. Its `RECEIPT_STATUSES` and `SCHEDULE_RUN_STATUSES` were already correct and already tested, so the sixth criterion needed no code.
- No client-side `"ok"` fallback. It would be redundant with the migration and has no live trigger.
- No `turn_stopped` handling in the reconciler. Scheduled runs call `run_turn` with no cancel control, so that path always emits `turn_end` with `state: "stopped"` instead. Handling it would be dead code here.
- No duration-display change for in-progress runs. Out of scope.

## Still open

- The reconciler now does one transcript file read per orphaned row inside the startup write transaction. At realistic orphan counts this is negligible, but it is more work under the lock than before. Noting it rather than pre-optimizing.
- The fourth criterion did not name a concrete failure mode. It is implemented as "a crash between the transcript's terminal write and the runs-table update must not downgrade a known outcome to interrupted." Say so if a different reading was intended.

Written boundary: this branch touched only `store.py` schedule/receipt paths, the schedule normalizers in `api.ts`, `ScheduledPage.tsx`, schedule CSS, and the schedule tests. No file owned by a parallel Codex lane was modified.

### DIFF
--- a/desktop/coworker/store.py
+++ b/desktop/coworker/store.py
@@ -237,6 +237,12 @@ class ConversationStore:
                 waiting_approval_count = COALESCE(waiting_approval_count, 0)
             """
         )
+        # Pre-S4 scheduler writes stored run_turn's raw "ok" status directly;
+        # RECEIPT_STATUSES has mapped it to "success" ever since, but rows
+        # written before that mapping existed are still on disk with the old
+        # text. Normalize them once so every consumer -- API, UI, future
+        # exports -- agrees on the shared status contract.
+        self._conn.execute("UPDATE runs SET status = 'success' WHERE status = 'ok'")
         try:
             self._conn.execute(
                 "ALTER TABLE sessions ADD COLUMN pinned INTEGER NOT NULL DEFAULT 0"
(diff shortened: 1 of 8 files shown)


FisherXZ (commented):
## Verdict
**Fix, then ship.**

## Summary
Legacy `ok` rows migrate to `success` on store init, so a completed sourcing review no longer renders as Failed. A status the client does not recognize becomes `unknown` and reads as Needs review. Crash-orphaned `running` rows now take their outcome from the scheduled session transcript when the last event is terminal.

The transcript is not per run. Every weekly fire reuses `sched-{job_id}`. After one completed week, a crash between `start_run` and the next `turn_start` copies last week's Completed receipt onto this week's row.

## Map
```mermaid
flowchart TD
  restart[Sidecar restart] --> reconcile[Reconcile rows still marked running]
  reconcile --> load["Load events for sched-jobId"]
  load --> last{Last event is turn_end or error?}
  last -->|yes| stamp[Copy that outcome onto this running row]
  last -->|no| interrupted[Stamp interrupted]
  prior["Previous week's turn_end still last"] -.->|session file is reused| stamp
```

Why Fix, then ship: #40's `ok` migration and the client `unknown` path hold. I reran the suites on `c40ad46`: pytest 732 passed, 3 skipped; GUI 385 passed across 43 files. Transcript agreement is the right seam. The last-event predicate is not scoped to this attempt.

## Blocking
1. A weekly routine that already completed once, then crashes after `start_run` and before the next `turn_start`, stores this week's receipt as Completed with last week's result. Scheduled jobs reuse `sched-{job_id}`. `_terminal_transcript_outcome` uses the file's last terminal event, so it invents an outcome the current attempt never produced. I reproduced it on this head: week 1 finished with "Week 1 found three contacts."; week 2 was still `running` with `duration_ms = None`; after reopen, run 2 was `success` / `duration_ms = 0` with week 1's text. #40 requires receipt status to agree with this run's terminal event, not a prior week's. The scheduler already snapshots `event_offset` for that reuse; the reconciler does not.

GitHub rejects `REQUEST_CHANGES` on the author's own PR, so this is posted as a comment review.

FisherXZ (commented):
## Verdict
**Ship.**

## Summary
Round-one leftover is closed. `start_run` now stores `event_offset` on the reused `sched-{job_id}` file, and the reconciler only reads events after that offset. A crash after `start_run` and before this week's `turn_start` stores Interrupted, not last week's Completed.

Loaded #40, QA T6, merge-base `418cd54...ae5d27d`, and the round-one review on `c40ad46`. Reproduced the leftover on this head: week 1 stayed Completed; week 2 was Interrupted. A week-2 `turn_end` still heals to Completed with this week's text.

pytest: 733 passed, 3 skipped in 35.43s. GUI: 385 passed across 43 files.

## Map
```mermaid
flowchart TD
  restart[Sidecar restart] --> reconcile[Reconcile rows still marked running]
  reconcile --> slice[Load events after this run's event_offset]
  slice --> last{Last remaining event is turn_end or error?}
  last -->|yes| stamp[Copy that outcome onto this running row]
  last -->|no| interrupted[Stamp interrupted]
  prior[Previous week's turn_end] -->|offset skips these events| slice
```

Why Ship: #40 requires receipt status to agree with this run's terminal event. The offset is the right seam for a reused session file. The `ok` migration and client `unknown` path from round one still hold.

## Blocking
None.

The failure I was most afraid of was week 2 still copying week 1, or the offset skipping this week's `turn_end`. Both paths were run on `ae5d27d`. They held.

PR #103, opened 2026-08-28
feat(desktop): add Sourcecado Doctor and the versioned migration registry

FisherXZ:
Partially addresses #73. The run-ownership checks are deliberately deferred; see "Not in this PR".

## Domain words

- **Durable store** — any file Sourcecado's state survives a restart in: a SQLite database, a JSONL append-only log, a JSON document, or an opaque file.
- **Migration registry** — one table listing every durable store, its declared version, and how to upgrade it.
- **Doctor** — a local diagnostic command that inspects those stores and reports problems.
- **Repair** — a change Doctor makes to fix a problem. Off by default.
- **Canary** — a fake secret planted in a test fixture to prove it never reaches output.

## Problem

Sourcecado had no release versioning. Stores were created if missing and columns were added by checking whether they already existed. That tells you nothing about which version a file on disk actually is, so packaging, updates, and rollback have nothing to check against.

## What this adds

`coworker/migrations.py` — one registry across every durable store. Each store declares a version and an upgrade path. `coworker/doctor.py` — checks, a bounded redacted report, backup, repair, and a CLI.

Doctor starts no model and performs no connector action. Dry-run is the default and prints the exact proposed repair with affected record counts. Repair writes a timestamped backup first. Only deterministic repairs run automatically.

An unknown future version fails closed and changes nothing.

## Measurements

```
799 passed, 3 skipped, 1 warning in 39.14s
```

Baseline before this branch was 729 passed, 3 skipped. This adds 64 tests in `test_migrations.py` (27) and `test_doctor.py` (37).

Re-run independently by the reviewing session on the pushed commit.

## Two test corrections, stated plainly

Both were caught during this work and are recorded so nobody later reads the diff and wonders whether a test was weakened.

**Rollback test.** The first version asserted byte-identity of the database file after a rollback. That is the wrong assertion. SQLite's online backup rewrites page layout, so a logically identical database can have a different byte stream. It now asserts `PRAGMA integrity_check` plus full row-content comparison of every table in the fixture. That is a stronger test, not a weaker one.

**Redaction coverage.** An audit during this work found the secret canaries covered SQLite, both JSONL kinds, and opaque files, but not the JSON document stores (`workspace_grants`, `shell_tasks`, `host_command_approvals`, `directory_requests`, `mcp.json`) or the memory directory. The claim "every store type is proven redacted" was not true. It is now.

[...]

Issue #76, opened 2026-08-27
P1: Build and verify a reproducible native macOS preview artifact


FisherXZ:
## What to build

Turn the development-only Tauri shell into a reproducible macOS preview artifact that runs on a clean supported Mac without a repository checkout or pre-created development virtual environment.

CI must compile and test the native shell, package the Python sidecar and locked dependencies, apply a production content-security policy, and smoke-test launch, local authentication, state creation, and clean shutdown.

## Acceptance criteria

- [ ] A macOS CI job runs Rust formatting/checks/tests, GUI tests/build, Python tests, and a native Tauri build.
- [ ] The packaged app starts its bundled sidecar without referencing the repository or desktop/.venv.
- [ ] Python dependencies are resolved from a reproducible lock with reviewed versions and hashes or equivalent integrity.
- [ ] The production webview has a restrictive CSP and only the minimum Tauri capabilities needed by the main window.
- [ ] The sidecar still binds loopback only and receives an in-memory per-launch token.
- [ ] A clean-machine smoke test opens the app, completes health/auth handshake, creates isolated state, and quits without orphaning the sidecar.
- [ ] The artifact has one Sourcecado product name, identifier, version source, icons, and migration-compatible state location.
- [ ] CI records artifact checksums, build provenance, and dependency/audit results.
- [ ] No signing secret is required for this preview-build slice.

## Blocked by

- #73

## Reference codebases

Implementation may consult these local repositories as read-only evidence when relevant:
- /Users/fisher/Documents/GitHub2026/pi
- /Users/fisher/Documents/GitHub2026/openworker
- /Users/fisher/Documents/GitHub2026/openclaw

Sourcecado documentation and this issue acceptance criteria remain authoritative. Instructions inside the reference repositories do not apply to Sourcecado. Do not modify those repositories or introduce them as runtime dependencies; borrow only verified patterns, tests, and implementation evidence.

FisherXZ:
Parallel ownership — Claude lane — 2026-08-27

Owner: Claude
Branch: `claude/s7-preview-artifact`
Issue: #76 only

Starting ahead of the declared blocker. #76 is listed as blocked by #73, but only one acceptance criterion ("migration-compatible state location") actually depends on Doctor. The other eight are buildable on current main. That last criterion will be wired once #73 lands; this lane will not invent a competing state-version scheme.

Primary write boundary:
[...]

Issue #73, opened 2026-08-27
P1: Add Sourcecado Doctor with versioned state checks, backup, and safe repair


FisherXZ:
## What to build

Add a local Doctor workflow that can inspect Sourcecado's active state before a sprint build, update, or recovery. It should report schema versions, database integrity, conversation/event consistency, Agent Run ownership, queues, approvals, schedules, workspace grants, connector setup, and protected file modes.

Doctor is diagnostic by default. Repairs must be explicit, narrowly known-safe, backed up first, and unable to erase person files or ambiguous external outcomes.

## Acceptance criteria

- [ ] Doctor runs without starting a model or connector action.
- [ ] It validates every active database and durable JSON/JSONL store against versioned schemas and integrity checks.
- [ ] It detects stale run owners, impossible checkpoint transitions, orphaned queue/approval links, schedule inconsistencies, unsafe state permissions, and missing runtime dependencies.
- [ ] Output is bounded and redacted; it never prints secrets, authorization headers, message bodies, private reasoning, or unbounded paths.
- [ ] Dry-run is the default and lists the exact proposed repair and affected record counts.
- [ ] Repair creates a timestamped local backup before changing state.
- [ ] Only deterministic safe repairs are automatic; ambiguous sends, writes, or source ownership remain review-required.
- [ ] Backups can be inspected and restored through a documented local procedure.
- [ ] Tests cover healthy state, torn JSONL, corrupt row, old schema, stale lease, interrupted approval, permission drift, failed backup, idempotent repair, and post-repair restart.

## Blocked by

- #63

## Sprint 5 versioned migration contract

Authoritative plan: [Sourcecado Sourcing OS Release Program — Execution Plan](https://app.notion.com/p/3c9d64acf06681af91dbee2d65715f70)

Doctor, packaging, updates, and rollback use one explicit migration registry across every active SQLite and durable JSON/JSONL store. Ad hoc create-if-missing or column-existence checks are not sufficient release versioning.

### Additional release acceptance criteria

- [ ] Every durable store reports a current schema/data version.
- [ ] Each supported prior version has a production-shaped upgrade fixture.
- [ ] Migration backup, failure, rollback, rerun, and post-restart behavior are tested.
- [ ] Unknown future versions fail closed without modifying state.
- [ ] The packaging and updater workflows consume the same compatibility contract.

## Reference codebases

Implementation may consult these local repositories as read-only evidence when relevant:
[...]

PR #104, opened 2026-08-28
feat(desktop): package a reproducible macOS preview artifact

FisherXZ:
Addresses #76. Three of nine criteria are partial; the table says which and why.

## Domain words

- **Sidecar** — the Python FastAPI process the desktop shell starts and talks to over loopback.
- **Shell** — the Rust/Tauri application that owns the window and spawns the sidecar.
- **onedir / onefile** — PyInstaller's two output shapes: a directory of files, or a single self-extracting executable.
- **Lock file** — a dependency list pinned to exact versions with hashes.

## Problem

The Tauri shell only ran from a repo checkout. It spawned `desktop/.venv/bin/python -m coworker.run` from a path computed off `CARGO_MANIFEST_DIR`, in every build mode. There was no content security policy, no dependency lock, and the product called itself three different things.

## Before and after

| | Before | After |
|---|---|---|
| Sidecar source | `desktop/.venv` in the repo | bundled resource inside the `.app` |
| CSP | `null` | restrictive policy |
| Tauri capabilities | window show/hide/focus/unminimize + `dialog:allow-open` | `core:default` |
| Product name | "Club" | "Sourcecado" |
| Bundle identifier | `com.codeology.club` | `com.codeology.sourcecado` |
| Version | three independent `0.0.1` literals | one, from `Cargo.toml` |
| Python deps | unpinned, no hashes | `requirements.lock`, hash-pinned |

## A real bug found while packaging

The first approach used PyInstaller `--onefile` shipped as a Tauri `externalBin`. That orphans the sidecar.

A onefile build's macOS bootloader forks a second process to run the real interpreter and returns the bootloader's PID to the caller. The shell's `Child::kill()` on quit kills that PID. The real sidecar keeps running. Reproduced twice with `pgrep` and process-tree inspection.

`--onedir` has no extraction step and no fork, so the PID Tauri spawns is the PID that answers requests and the one `kill()` stops. onedir ships as a directory, so it goes through `bundle.resources` rather than `externalBin`, resolved at runtime with `app.path().resolve(...)` under `BaseDirectory::Resource`.

This is the reason for the packaging shape, not a stylistic choice.

## Measurements

Every command below was executed locally on macOS `aarch64-apple-darwin`.

```
cargo fmt --check                     clean
cargo clippy --all-targets -D warnings clean, zero warnings
cargo test                            ok. 0 passed; 0 failed (no Rust unit surface)
npm test                              43 files, 382 tests passed
npm run build                         typecheck + vite build succeeded
pytest -q                             733 passed, 3 skipped
build_sidecar.sh                      53MB onedir tree
pip-audit -r requirements.lock        No known vulnerabilities found
npm run tauri -- build                release compile ~23s, Sourcecado.app bundled
```

The Python suite was re-run independently by the reviewing session on the pushed commit: 733 passed, 3 skipped.

Smoke test, run against both the standalone frozen sidecar and the copy inside the built `.app`: health handshake returns 200 with the expected payload, unauthenticated request returns 401, isolated state created under a per-run temporary `CLUB_STATE_DIR`, clean kill with zero orphaned processes confirmed by `pgrep`.

`codesign -dv` on the built app confirms an adhoc signature with no identity, so no signing secret is needed.

## CI run

The `macOS preview artifact` job ran on a clean GitHub-hosted macOS runner and passed in 5m51s. Steps executed, in order:

```
Install Rust components
Install Python dependencies      (from requirements.lock)
Python tests                     841 passed, 3 skipped
Install GUI dependencies
GUI tests                        45 files, 404 passed
GUI typecheck and build
Rust format check                cargo fmt --check
Build sidecar
Rust lint                        cargo clippy --all-targets -- -D warnings
Rust tests
Native Tauri build               Bundling Sourcecado.app / Finished 1 bundle
Smoke test the packaged sidecar
Audit locked Python dependencies No known vulnerabilities found
Record build provenance and checksums
upload-artifact
```

This matters beyond ticking criterion 1. The runner is a clean machine with no repository checkout state and no developer virtual environment, which is exactly the condition criterion 2 describes. The packaged app building and smoke-testing there is stronger evidence than any local run.

The Python count is 841 rather than the 733 measured locally because CI tests the merge with main, which has moved ahead.

## Cold rebuild

`target/` and the generated `resources/` were deleted and the whole thing rebuilt from scratch, every Rust crate recompiled:

```
Finished release profile [optimized] target(s) in 1m 29s
Bundling Sourcecado.app
Finished 1 bundle
```

Two successful native builds this session, the second fully cold, same result both times.

Checksums on the cold-built artifact:

```
d7dc9c930588ad4d9f193dd98beb920e4f47930201788c953f215e70fbb6d6ce  Contents/MacOS/sourcecado-desktop
3805423e2a6891188999021c18a8ae431c67611b4936b3a2daba63e6f9989d19  Contents/Resources/resources/sourcecado-sidecar/sourcecado-sidecar
```

## Window-level launch

The freshly built app was launched with an isolated `CLUB_STATE_DIR`:

- Shell came up as PID 84341 and spawned its sidecar as a direct child, PID 84411, from `Contents/Resources/resources/sourcecado-sidecar/`. The resource path resolves in a live launch, not only on disk.
- `lsof -p 84411` showed exactly one listening socket: `TCP 127.0.0.1:8765 (LISTEN)`. Loopback only, verified live.
- The per-launch token was read from the sidecar's own environment and used externally: `/v1/health` with the header returned 200 and the expected payload; without it, 401.
- `sidecar.log` shows real frontend traffic, not synthetic requests: `GET /v1/inbox`, `/v1/sessions`, `/v1/persona`, an accepted `WebSocket /ws/chat`, and `PATCH /v1/navigation`. The webview loaded, authenticated, and used both HTTP and the chat socket.
- Isolated state was created by actual app usage: `club.db`, `people.db`, `mcp.json`, and a real conversation `.jsonl`, all under the per-run temporary state directory.

Both shutdown paths were confirmed independently:

| Path | How | Result |
|---|---|---|
| Shell's own `child.kill()` on Quit | `osascript ... to quit` | Both PIDs gone, `pgrep` empty |
| Sidecar's `_exit_when_orphaned` watchdog | `kill -TERM` on the shell, measured by the reviewing session against the installed `/Applications` copy | Process exits, sidecar count zero |

The first exercises `RunEvent::Exit`. The second exercises the watchdog for an abrupt parent death. Both are clean.

## Acceptance criteria

| # | Criterion | Status | Evidence |
|---|---|---|---|
| 1 | macOS CI job runs Rust, GUI, Python, native build | Done | Job executed on a clean GitHub-hosted macOS runner and passed in 5m51s. See "CI run" below. |
| 2 | Packaged app starts bundled sidecar without repo or `.venv` | Done | Release path resolves only a bundled resource. Now also built and smoke-tested on a clean CI runner with no developer venv, which is the criterion's actual condition. |
| 3 | Deps from a reproducible lock with hashes | Done | `requirements.lock` installs with `--require-hashes`; `pip-audit` clean |
| 4 | Restrictive CSP, minimum capabilities | Done | CSP set; capabilities trimmed after a full-repo search found zero JS call sites needing the removed ones |
| 5 | Sidecar loopback-only with per-launch token | Done | Guard code untouched; 4 regression tests added |
| 6 | Smoke test: open, handshake, isolated state, quit, no orphan | Done | Full window-level launch verified first-hand; both shutdown paths confirmed clean. See "Window-level launch" below. |
| 7 | One name, identifier, version, icons, state location | Partial | Name/identifier/version unified and confirmed in the built `Info.plist`; icons are still Tauri's placeholder artwork |
| 8 | CI records checksums, provenance, audit results | Done | `Record build provenance and checksums` and `Audit locked Python dependencies` both ran in CI; artifact uploaded. |
| 9 | No signing secret required | Done | `codesign -dv` confirms adhoc only |

## Not in this PR

- Code signing, notarization, and `.dmg`. `.dmg` creation fails in this environment: `bundle_dmg.sh`'s AppleScript step errors with Finder `-10006`, a known non-interactive-session failure. No criterion requires a `.dmg`.
- Real Sourcecado icon artwork. No source asset was available. The icon *configuration* is consistent; the images are placeholders.
- Intel `x86_64-apple-darwin`. Built and tested on Apple silicon only.
- `cargo audit` / `cargo deny`. Criterion 3 is about the Python lock.

## Corrections to the ticket's premises

The brief given to this lane said CI does not run the desktop pytest suite. That is stale. `.github/workflows/ci.yml` already had a `sidecar` job running `pytest -q` on Ubuntu. It ran from the unpinned `requirements.txt`; this PR switches that one line to the lock.

## Still open

- `cargo build` and `tauri dev` now require `resources/sourcecado-sidecar/` to exist at compile time, because Tauri's build script validates `bundle.resources` paths unconditionally rather than only when bundling. This is a new local prerequisite for any Rust build in this crate. Documented in `desktop/docs/packaging.md`, but it is a real workflow change for anyone working in `src-tauri`.
- The onedir sidecar takes roughly 4 to 7 seconds to answer its first request while the interpreter warms up. Not addressed.
- One-line `.PHONY:` conflict in `Makefile` against #73's branch, whichever merges second.
- The release-path `sidecar_command` never calls `.current_dir()`, so the sidecar inherits whatever working directory the parent had at spawn. Harmless today: nothing in `coworker` reads paths relative to cwd (personas and skills resolve from `__file__`, state comes from the explicit `CLUB_STATE_DIR`). Naming it rather than leaving it implicit. A user launching from Finder gets `/` or their home directory.
- `src-tauri/binaries/` does not exist and was never committed. The sidecar ships from `src-tauri/resources/sourcecado-sidecar/`, which is generated and gitignored.

## Coupling to #73

`state_dir()` still resolves to `CLUB_STATE_DIR` or `~/.config/club`, reused verbatim. No schema-version marker or migration hook was invented here. #73's migration registry (PR #103) should run against this same path.

Write boundary: `src-tauri/`, CI, the dependency lock, new packaging scripts and tests, docs, and Makefile targets. No file under `desktop/coworker/` was modified.


FisherXZ:
CI update: the `macOS preview artifact` job has now run on a clean GitHub-hosted macOS runner and passed in 5m51s ([run](https://github.com/FisherXZ/Sourcecado/actions/runs/33130670925/job/98719184218)).

That closes the two criteria this PR originally listed as partial:

- Criterion 1 — the job itself executed, not just its constituent commands locally.
- Criterion 8 — `Record build provenance and checksums` and `Audit locked Python dependencies` both ran; the artifact uploaded.

It also strengthens criterion 2. The runner has no repository checkout state and no developer virtual environment, which is the criterion's actual condition. The app built and smoke-tested there.

Remaining partial: criterion 7, because the icons are still Tauri's placeholder artwork. No source asset was available. Everything else in criterion 7 (one name, identifier, version source, state location) is done and confirmed in the built `Info.plist`.

Body updated to match.

desktop/surfaces/gui/src-tauri/capabilities/default.json
```diff
@@ -1,13 +1,6 @@
...
   "windows": ["main"],
-  "permissions": [
-    "core:default",
-    "core:window:allow-hide",
-    "core:window:allow-show",
-    "core:window:allow-set-focus",
-    "core:window:allow-unminimize",
-    "dialog:allow-open"
-  ]
+  "permissions": ["core:default"]
```
FisherXZ: Workspace Settings still calls `window.__TAURI__.dialog.open` (`pickDirectory` in `api.ts`). The dialog plugin IIFE maps that to `plugin:dialog|open`. `core:default` does not grant it.

```suggestion
  "permissions": ["core:default", "dialog:allow-open"]
```

FisherXZ (commented):
Round 2 on 6a5b411. Loaded issue #76, the merge-base diff against origin/main (418cd54), and the round-one miss on dialog:allow-open.

The main window now grants dialog:allow-open next to core:default. pickDirectory is still the only JS Tauri call, and plugin-dialog maps it to plugin:dialog|open. The new workspaceApi test reads that capability file.

Local suites on this head: desktop pytest 733 passed, 3 skipped; GUI vitest 43 files, 383 passed, including the dialog:allow-open grant. GitHub Actions on this SHA: Python sidecar, Desktop GUI, and macOS preview artifact all succeeded, including the packaged-sidecar smoke test. That held.

PR #108, opened 2026-08-28
feat(desktop): inspect agent runs through one unified ledger

### LINKED ISSUE #74: P1: Inspect ordinary agent runs through one unified local ledger

FisherXZ:
Closes #74.

## Domain words

- **Receipt** — one Agent Run rendered as operational evidence: what ran, what it touched, how it ended.
- **Owner-native receipt** — the record kept by whatever system actually holds the authority, e.g. the Inbox item for an approval. The source of truth for whether a human said yes.
- **Facet** — one section of a receipt: approvals, sources, artifacts, usage, model attempts, outcome.
- **Checkpoint sequence** — the dense 1..N numbering the run store writes, which makes it possible to tell whether a record covers a run's whole life.

## Problem

The durable Agent Run landed in #63, but there was no way to look at one. An operator could not ask what a run did, which sources it touched, what it asked permission for, or how it ended.

## What this adds

Four modules, read-only:

| Module | Holds |
|---|---|
| `run_evidence.py` | Evidence vocabulary and record analysis |
| `run_receipt.py` | The receipt projection. Pure — no store, no clock, no network |
| `run_ledger.py` | Reader over the run store, query, retention |
| `run_ledger_api.py` | HTTP surface. **No write verb** |

Not to be confused with the existing `coworker/ledger.py`, which maps tool results into person-file events. That is the person timeline; this is the run receipt. It is untouched.

## Keeping it from becoming a second transcript

The rule is mechanical rather than aspirational:

> A receipt renders only fields `run_receipt.py` names, and every named field is an identifier, an enum, a count, a timestamp, or a bounded reason the runtime itself wrote.

Message bodies, tool arguments, command output, prompts, raw provider errors, credentials, and model reasoning **have no field to land in**. `_pick` copies named scalars and never a nested structure, so a payload cannot ride out whole.

This is a second allowlist, independent of the write-time one in `agent_runs.CHECKPOINT_PAYLOAD_FIELDS`. Both must hold on their own. A read side that leaned on the write side would keep passing its tests after the write side changed — which is exactly the kind of coupling that turns into a leak six months later.

Free text appears in exactly three bounded places, all already through `redact_secrets` at write time: `reason`, `error_summary`, and reference titles. The ledger adds no free-text field of its own.

## Criterion 6: silence is not absence

Every facet carries one value from a closed vocabulary:

| Value | Means |
|---|---|
| `present` | The record says so directly |
| `absent` | It positively did not happen |
| `partial` | Some settled, some did not |
| `missing` | We do not know — the record has a hole, or the run is open |
| `ambiguous` | The run knows it does not know: an external effect never reported back |
| `unsupported` | This build cannot express or verify the fact. Never "it did not happen" |
| `expired` | The evidence existed and retention aged it out on purpose |

"No approval was requested" is `absent`. "We do not know whether an approval was requested" is `missing`. Different operator situations; they must never render the same.

What separates them is a computed property, not a guess. A facet with no evidence is `absent` only when the run is terminal, **and** the stored checkpoints are contiguous ending at `checkpoint_sequence`, **and** no checkpoint kind is unknown to this build, **and** no checkpoint is `process_interrupted` or `tool_outcome_unknown`. Otherwise it is `unsupported`, `missing`, or `expired`, in that precedence.

An unreadable record withholds a conclusion even where entries exist. The entries stay visible, but the section reads `unsupported` — because a checkpoint this build cannot interpret might have been another tool call, another approval, or another model attempt.

## Criterion 7: a ledger row is not authority

The run store records which approval ids a run touched. **It never records whether the owner said yes.**

Every decision is resolved against the owner-native record. An approval id the Inbox does not know renders `missing`, never an allowance. Pending renders `partial`. Expired renders `expired` and carries no decision. Only a resolved owner-native item with `allow` or `deny` produces a decision field at all, and an authorized action whose execution was interrupted renders `ambiguous`.

Three structures hold this up:

1. `RunLedger`'s only write is `prune_checkpoints`, which deletes checkpoint rows and touches `agent_runs` never. It holds no lease and cannot acquire one.
2. `run_ledger_api` has no write verb — no route creates a run, changes a state, or records a decision.
3. `_approval_entry` sources `decision`, `state`, and `actor` only from the owner-native record.

## Measurements

```
1048 passed, 3 skipped, 1 warning in 43.87s
```

## Edits to code merged in #63

Additive only, and small: `agent_runs.py` +6 lines, `server.py` +8 lines. The compare-and-swap write path and the lease semantics from #105 are untouched — those are load-bearing correctness that already shipped.

## Not in this PR

- No new GUI surface. Criterion 4's entry points are proven at the API layer; adding UI would have meant touching files parallel lanes hold.
- Retention prunes checkpoint detail only. Row-backed facets — sources, artifacts, usage, approvals, outcome — survive pruning, because the row is written in the same transaction as every checkpoint. Pruning turns a checkpoint-backed facet `expired` and leaves a row-backed one alone.

## Still open

- Retention is bounded by checkpoint pruning, not by run-row deletion. Runs accumulate. That is deliberate for now — the rows are small and referenced person/source/artifact identities must survive — but it is a growth curve worth watching.

### DIFF
--- a/desktop/coworker/agent_run_repository.py
+++ b/desktop/coworker/agent_run_repository.py
@@ -32,6 +32,8 @@ from coworker.agent_run_state import (
     validate_transition,
 )
 from coworker.agent_runs import (
+    CHECKPOINT_KINDS,
+    INCOMPLETE_RECORD_KINDS,
     RUN_TRIGGERS,
     add_usage,
     checkpoint_payload,
@@ -185,7 +187,11 @@ class AgentRunRepository:
         self,
         *,
         session_id: str | None = None,
+        person_id: str | None = None,
+        trigger: str | None = None,
         states: Iterable[str] | None = None,
+        created_after: datetime | None = None,
+        created_before: datetime | None = None,
         limit: int = 100,
     ) -> list[dict[str, Any]]:
         clauses: list[str] = []
@@ -193,6 +199,18 @@ class AgentRunRepository:
         if session_id is not None:
             clauses.append("session_id = ?")
             params.append(str(session_id))
+        if person_id is not None:
+            clauses.append("person_id = ?")
+            params.append(str(person_id))
+        if trigger is not None:
+            clauses.append("trigger = ?")
+            params.append(str(trigger))
+        if created_after is not None:
+            clauses.append("created_at >= ?")
+            params.append(_stamp(created_after))
+        if created_before is not None:
+            clauses.append("created_at <= ?")
+            params.append(_stamp(created_before))
         wanted = tuple(states or ())
         if wanted:
             clauses.append(f"current_state IN ({','.join('?' * len(wanted))})")
@@ -406,6 +424,48 @@ class AgentRunRepository:
                 )
             ),
         )
+
+    # --- retention -------------------------------------------------------
+
+    def prune_checkpoints(self, *, finished_before: datetime) -> list[str]:
+        """Drop step detail for runs that finished before the cutoff.
+
+        This is the only write in the store that is not a compare-and-swap
+        against a lease, and it is deliberately narrow: it deletes rows from
+        `agent_run_checkpoints` and touches `agent_runs` never. Run identity,
+        person, source and artifact references, usage, approvals, and outcome
+        all live on the run row and survive. A terminal run never moves again,
+        so no lease can be racing this.
+
+        A run whose record marks a hole, or carries a checkpoint kind this
+        build cannot read, is skipped: retention must not delete the evidence
+        that evidence is missing.
+        """
+        cutoff = _stamp(finished_before)
+        readable = tuple(sorted(CHECKPOINT_KINDS - INCOMPLETE_RECORD_KINDS))
+        pruned: list[str] = []
+        with self._write():
+            rows = self._conn.execute(
+                f"""
+                SELECT run_id FROM agent_runs
+                WHERE finished_at IS NOT NULL AND finished_at < ?
+                  AND NOT EXISTS (
+                      SELECT 1 FROM agent_run_checkpoints c
+                      WHERE c.run_id = agent_runs.run_id
+                        AND c.kind NOT IN ({','.join('?' * len(readable))})
+                  )
+                ORDER BY finished_at
+                """,
+                (cutoff, *readable),
+            ).fetchall()
+            for row in rows:
+                run_id = str(row["run_id"])
+                removed = self._conn.execute(
+                    "DELETE FROM agent_run_checkpoints WHERE run_id = ?", (run_id,)
+                ).rowcount
+                if removed:
+                    pruned.append(run_id)
+        return pruned
 
     # --- recovery --------------------------------------------------------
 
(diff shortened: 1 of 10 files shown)


FisherXZ (commented):
## Verdict
**Ship.**

## Summary
This PR makes the durable Agent Run from #63 inspectable. One read-only receipt covers chat, queued, scheduled, resumed, and failed runs.

The projection is its own allowlist. It is independent of the write-time checkpoint allowlist. Message bodies, tool arguments, credentials, and model reasoning have no field to land in.

Silence is not absence. "No approval was requested" and "we do not know whether one was" are different evidence values. A ledger row is not authority. A run-store status of allow never becomes a decision. Only the Inbox item does. The HTTP surface has no write verb.

## Map
```mermaid
flowchart TD
  Op[Operator] --> API["GET /v1/agent-runs or /v1/agent-runs/id"]
  API --> Ledger[RunLedger]
  Ledger --> Rows[Run store row plus checkpoints]
  Ledger --> Inbox[Inbox get_inbox]
  Rows --> Analyze[analyze_record]
  Analyze --> Receipt["build_receipt named scalars only"]
  Inbox --> Decision["_approval_entry"]
  Decision --> Receipt
  Rows -.->|payload status never a decision| Decision
  Receipt --> JSON["JSON no-store"]
```

Why Ship: Issue #74 asks for one inspectable receipt that is evidence, not a transcript, and not authority. I loaded #74, the PR body, `desktop/docs/run-ledger.md`, and the merge-base diff against `a2cef0e`. I ran `cd desktop && .venv/bin/pytest -q` on `9609062`: 1048 passed, 3 skipped, 1 warning in 42.13s. The 30 new ledger tests passed in 0.52s.

## Blocking
None.

The failure I was most afraid of was a receipt echoing planted tool arguments, or treating a run-store allow as permission. Both paths are tested on this head and they passed.

Issue #66, opened 2026-08-27
P0: Prove reviewed enrich, draft, and Gmail send for one person


FisherXZ:
## What to build

Close the highest-stakes sourcing path for one bound person: deliberate Apollo enrichment when needed, a tailored Gmail draft, explicit review, and approved send from Sourcecado. Every costly or external step must have one immutable authority chain and a durable outcome.

Automated tests use fakes. Final acceptance includes one controlled live canary chosen and approved by Fisher.

## Acceptance criteria

- [ ] The flow starts from a person-bound sourcing chat and never guesses the recipient.
- [ ] Apollo enrichment shows the exact person and credit-spend reason before approval.
- [ ] Enrichment updates only the approved person and records the Apollo source receipt.
- [ ] The Gmail draft is editable and visibly not sent before send approval.
- [ ] Send approval binds the exact Gmail account, draft id, recipient, subject, and reviewed body version.
- [ ] Allow executes at most once; deny, cancel, expiry, stale draft, and failed submission send nothing.
- [ ] A provider retry, transport reconnect, or duplicate approval submission never automatically repeats the send; process-crash ambiguity is hardened later in #63.
- [ ] Successful send records Gmail message and thread identity, advances the person to Open if needed, and leaves an inspectable person/run receipt.
- [ ] Fake-service integration tests cover every approval, retry, reconnect, and duplicate-submission boundary.
- [ ] Fisher performs and approves one controlled live canary without exposing credentials in logs or issue comments.

## Blocked by

- #62
- #54
- #55
- #59
- #71
- #80

## Reference codebases

Implementation may consult these local repositories as read-only evidence when relevant:
- /Users/fisher/Documents/GitHub2026/pi
- /Users/fisher/Documents/GitHub2026/openworker
- /Users/fisher/Documents/GitHub2026/openclaw

Sourcecado documentation and this issue acceptance criteria remain authoritative. Instructions inside the reference repositories do not apply to Sourcecado. Do not modify those repositories or introduce them as runtime dependencies; borrow only verified patterns, tests, and implementation evidence.

FisherXZ:
Parallel ownership — Claude lane — 2026-08-27

Owner: Claude
Branch: `claude/s3-approved-send`
Issue: #66 — the AFK portion

All declared blockers are resolved: #62, #54, #59, #71 closed; #80 closed; #55's implementation merged as PR #97.

[...]

PR #109, opened 2026-08-28
feat(desktop): bind and prove the approved Gmail send path

FisherXZ:
Addresses #66, criteria 1 through 9. **Criterion 10, the live canary, is not done and needs Fisher.** See the last section.

## Domain words

- **Claim** — the single database write that decides which caller is allowed to execute an approved action. Exactly one caller wins.
- **Reviewed version** — the exact draft body the director saw when they approved. Identified by a digest, not stored as content.
- **Attempt vs send** — two separate counters on the Gmail fake. Attempts counts every call into `send`, including ones that raise. Sends counts messages that actually left. This makes "we never reached Gmail" distinguishable from "Gmail refused us".

## Problem

The highest-stakes path in the product — enrich, draft, review, send — had no binding between what the director approved and what actually went out. A duplicate submission, a reconnect, or an edited draft could each have produced a send nobody authorized.

## At-most-once: four layers, three of them already existed

The important design decision here was **not** to build a new locking mechanism.

| Layer | Where | New? |
|---|---|---|
| The claim | `store.py:1000`, `decide_and_claim_inbox_execution` | No — reused, untouched |
| Terminal guard | `store.py:1089`, `complete_inbox_execution` | No — reused, untouched |
| Retry allowlist | `permissions.RETRY_SAFE` | No — reused, untouched |
| Authority check | `gmail.py`, `send_reviewed_draft` | Yes |

The claim records the decision and claims execution in one statement:

```sql
UPDATE inbox SET execution_status='executing'
WHERE id=? AND COALESCE(execution_status,'pending')='pending'
```

Exactly one caller sees `rowcount == 1`. Every other caller — a duplicate POST, a reconnected window, a second socket — gets `claimed=False` and is routed to `wait_for_execution`, which reads an outcome and cannot produce one. **The second submitter never reaches send code at all.**

`gmail_send` is in `ASK`, absent from `AUTO`, and absent from `RETRY_SAFE`, which is what makes a provider retry unable to replay a send. A test asserts all four facts so a future edit to those sets breaks loudly. `permissions.py` was deliberately not modified — the machinery was already correct.

The one hole per-approval claiming cannot see is two separate approvals for one draft. That is closed because Gmail deletes a draft on send, and `FakeGmail` now models that, so a second send of the same draft id is a 404.

## Boundary matrix

Every row measured on the fake's counters, not on a status string.

| Case | attempts | sends |
|---|---|---|
| allow once | 1 | 1 |
| deny | 0 | 0 |
| cancel before decision | 0 | 0 |
| expiry | 0 | 0 |
| body edited after review | 1 | 0 |
| recipient changed after review | 1 | 0 |
| subject changed after review | 1 | 0 |
| account changed | 1 | 0 |
| failed submission, then retry | 1 | 0 |
| duplicate submission ×3 | 1 | 1 |
| two concurrent submissions on a barrier | 1 | 1 |
| reconnect mid-send, then resubmit | 1 | 1 |
| two approvals, one draft | 1 | 1 |
| approval for an already-sent draft | 1 | 1 |
| draft addressed elsewhere | 0 | 0 |
| unbound or wrong-person chat | 0 | 0 |
| receipt write fails after a real send | 1 | 1 |

Apollo, counted on HTTP calls because credits are money: deny 0, expiry 0, duplicate submission 1, unmatched person 0, approved 1.

## Proving the negative tests are not vacuous

A test asserting "zero sends" is worthless if the flow never reached the point where a send was possible. Two independent mechanisms:

**In-test**, the signal is `execution_status`. `failed` means the claim was granted and the executor really ran, so zero sends is a guard working rather than a path nobody reached. Every negative test also runs a positive control on the identical untampered fixture and asserts it sends exactly once.

**Out-of-test**, a mutation harness breaks one guard at a time and requires the owning test to go red. **21 breaks, 21 red, zero vacuous.**

It found three things that would otherwise have shipped:

1. Two `already_sent` guards were dead code. Gmail deletes a sent draft, so `get_draft` 404s first and those branches were unreachable. Removed, so the tested mechanism is the real one.
2. The authority-mismatch tests did not assert that a refusal leaves no send receipt and no sequence advance. Injecting a phantom receipt into the refusal path *passed*. Fixed.
3. The harness itself had a stale-`.pyc` bug — same-second writes with identical size deltas let CPython reuse the previous mutation's bytecode — producing two false GREENs. Fixed by clearing `__pycache__` per run; both then went red.

That third one is worth naming. It is the same class of failure as the rebase hazard below: it looks clean and is wrong.

## Measurements

```
1082 passed, 3 skipped, 1 warning in 47.80s
 Test Files  47 passed (47)
      Tests  425 passed (425)
```

All re-run independently by the reviewing session on the pushed commit.

## Rebase note

Rebased onto `7563a77`. One conflict in `server.py`, a plain import collision; both kept.

A sibling lane found that git can place an identical closing line *outside* the conflict markers as though both sides shared it, so a naive both-sides-keep silently drops a brace without producing a conflict marker. Rather than trusting a careful read, this lane ran delimiter arithmetic per file: merged count must equal base + (mine − base) + (theirs − base) for each of `{ } ( ) [ ]`, and each file must be internally balanced. All five overlapping files passed.

Test arithmetic also closes exactly: base 919 + main's 129 new + this branch's 34 = 1082, which is what the tree gives. Nothing was lost on either side.

## Design decisions

- **Dedicated person-bound server routes over extending the model's `gmail_send` tool path**, because taking the recipient from the person file is a stronger guarantee than trusting model-generated arguments, and the park and receipt sites are in `turn.py`/`ledger.py`, outside this lane's boundary.
- **A body digest over the body itself** in the approval record, because a digest identifies the reviewed version without storing content.
- **A normalized digest over a raw one**, because Gmail round-trips bodies through quoted-printable and CRLF; a raw digest would false-positive on every live send.
- **`get_draft` always reads live, over the previous local-cache short-circuit**, because a cached copy hides an edit made in Gmail — the exact hazard the gate exists to catch.
- **Modelling Gmail's delete-on-send in the fake over a permissive fake**, because it makes a duplicate send loud instead of silent.
- **Reusing `decide_and_claim_inbox_execution` over a new send-specific lock**, because a second at-most-once mechanism is a second thing that can disagree with the first.

## NOT DONE: criterion 10, the live canary

This PR sends no real mail and touches no real credential. Criterion 10 requires a human. Steps, in order:

1. Connect Gmail via `/v1/gmail/connect` with an account you will send from. Confirm `/v1/gmail` reports `connected`.
2. Optionally set `APOLLO_API_KEY` to exercise enrichment. Without it the enrich route returns 409 cleanly.
3. Pick one real person — yourself or a colleague who expects it. Open the person file, **Create sourcing chat**, then in **Outreach**: compose, **Create Gmail draft**.
4. **The step that matters most.** Open the draft in Gmail, confirm it is unsent, edit one word, return, click **Re-read draft from Gmail**, and confirm the reviewed-version hash changes. This is the only exercise of the live `format=full` read path, which today has fake coverage only.
5. **Request approval to send**, then **Allow once and send**.
6. Confirm exactly one message in Sent. Check the person timeline shows one `send` event with the real Gmail `message_id` and `thread_id`, and that the person moved to Open.
7. One live negative worth doing: request approval, edit the draft in Gmail, then Allow. It must refuse with "The draft body changed after this send was approved" and send nothing.

## Still open

- **The live digest risk, and it is the real one.** The body digest depends on `_flatten_body`'s decoding matching what `create_draft` wrote. `normalize_body` handles CRLF and trailing whitespace, but a real quoted-printable round trip on long lines or non-ASCII could shift bytes and cause a false stale-draft refusal. It fails safe — nothing sends — but it would be an annoying false alarm, and step 4 above is the cheapest way to find out. Fakes cannot settle this.
- The reconnect test asserts a disjunction (200-idempotent or 202-pending) because the branch depends on timing against a 60s deadline. In practice it always takes 200. The send-count assertion is exact either way.
- `enrichment_match` will match on an obfuscated last name like `"L."`, wasting a credit on a bad query. Pre-existing behaviour of what `keep_from_apollo` stores, not introduced here. Worth a follow-up ticket.
- `record_approved_send` is idempotent on the Gmail message id, but if the receipt write fails *after* the send, the message is already gone. That path returns `ok: true` with a `receipt_error` and no person event, and has a test. Honest, but it leaves a sent message with no timeline entry.

## Not in this PR

- The chat-driven `gmail_send` tool path still has no body-version binding. Its park site is `turn.py`, outside this boundary.
- No `permission_required` chat event for the new bound approvals, so they surface in the person file rather than in chat. Wiring needs `turn.py` and the `protocol.ts` allowlist.
- No in-app draft editor. Gmail owns the draft; the panel re-reads it.
- No spend ceiling on Apollo, only per-call approval.

## Summary by CodeRabbit

* **New Features**
  * Added person-bound outreach drafting, review, approval, and Gmail sending.
  * Added safeguards that prevent sending modified, duplicate, or mismatched drafts.
  * Added Apollo contact enrichment with approval, credit-cost visibility, and source tracking.
  * Added outreach controls and status details to person profiles.
* **Bug Fixes**
  * Improved protection against duplicate sends and repeated enrichment charges.
* **Documentation**
  * Added a reference guide for the reviewed outreach and sending workflow.
* **Tests**
  * Added comprehensive coverage for outreach, approvals, enrichment, and delivery safeguards.

### DIFF
--- a/desktop/coworker/apollo.py
+++ b/desktop/coworker/apollo.py
@@ -13,6 +13,10 @@ MATCH_URL = "https://api.apollo.io/api/v1/people/match"
 
 MISSING_KEY = "APOLLO_API_KEY is not configured."
 
+# One people/match call spends one Apollo credit. The approval card states the
+# cost before the director allows it; nothing here decides to spend on its own.
+ENRICH_CREDIT_COST = 1
+
 
 class HttpError(RuntimeError):
     def __init__(self, status: int, url: str = "", body: object | None = None) -> None:
@@ -252,10 +256,63 @@ def enrich_contact(
     phones = person.get("phone_numbers") or []
     phone = phones[0].get("raw_number") if phones else None
     return {
+        "apolloId": person.get("id") or None,
         "name": person.get("name") or None,
         "title": person.get("title") or None,
         "organizationName": (person.get("organization") or {}).get("name") or None,
         "linkedinUrl": person.get("linkedin_url") or None,
         "email": person.get("email") or None,
         "phone": phone or None,
+    }
+
+
+def enrichment_match(person: dict[str, Any]) -> dict[str, Any] | None:
+    """The lookup Apollo will be asked for, taken from the person file.
+
+    Returns None when the person file cannot identify anybody. Apollo's match
+    endpoint needs an email, or a first and last name; guessing either one is
+    how the wrong contact gets enriched and the credit gets wasted.
+    """
+    email = str(person.get("email") or "").strip()
+    first = str(person.get("first_name") or "").strip()
+    last = str(person.get("last_name") or "").strip()
+    company = str(person.get("company") or "").strip()
+    if email:
+        return {"email": email, "organization_name": company or None}
+    if first and last:
+        return {
+            "first_name": first,
+            "last_name": last,
+            "organization_name": company or None,
+        }
+    return None
+
+
+def enrichment_resource(
+    person: dict[str, Any], match: dict[str, Any]
+) -> dict[str, Any]:
+    """The approval card for one enrichment: which person, and what it costs.
+
+    Names the exact person file being changed, not a search row, so Allow can
+    never land the facts on somebody else.
+    """
+    display = " ".join(
+        part
+        for part in (person.get("first_name"), person.get("last_name"))
+        if part
+    ).strip()
+    matched_on = "email" if match.get("email") else "name"
+    return {
+        "kind": "apollo_enrichment",
+        "person_id": str(person.get("person_id") or ""),
+        "person": display or "Unnamed person",
+        "title": person.get("title") or None,
+        "company": person.get("company") or None,
+        "matched_on": matched_on,
+        "credits": ENRICH_CREDIT_COST,
+        "reason": (
+            f"Spends {ENRICH_CREDIT_COST} Apollo credit to look up "
+            f"{display or 'this person'} by {matched_on} and write the result "
+            "to this person file only."
+        ),
     }
--- a/desktop/coworker/gmail.py
+++ b/desktop/coworker/gmail.py
@@ -1,12 +1,21 @@
 """Gmail draft client — injectable so tests never hit the network.
 
-Slice 12: Gmail API users.drafts.create after OAuth. There is no send method on
-the live client. FakeGmail stays for pytest. Secrets stay out of prompts.
+Slice 12: Gmail API users.drafts.create after OAuth. FakeGmail stays for
+pytest. Secrets stay out of prompts.
+
+Send is the one external effect in this module that costs a real message to a
+real person. It is guarded by a SendAuthority: the identity the director
+actually reviewed and approved. The authority is re-checked against the live
+draft immediately before the send call, so a draft edited after review sends
+nothing. The authority carries a digest of the reviewed body, never the body
+itself and never a token.
 """
 
 from __future__ import annotations
 
 import base64
+import hashlib
+from dataclasses import dataclass
 from email.message import EmailMessage
 from typing import Any, Protocol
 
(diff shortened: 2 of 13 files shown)


FisherXZ (commented):
## Verdict
**Ship.**

Loaded #66 criteria 1–9, the spring spec's approve-to-send lock, and the diff against merge-base `7563a77`. Re-ran `desktop/.venv/bin/pytest -q` (1082 passed, 3 skipped) and `desktop/surfaces/gui` `npm test` (47 files, 425 passed). The claim, live draft re-read, and FakeGmail delete-on-send held: one allow produced one send; deny, stale draft, duplicate, and reconnect produced no extra send.

## Blocking
None.

The failure I was most afraid of was a live Gmail `To` header like `Ada Lovelace <ada@analytic.example>` still matching the person file. On this head that path 409s `recipient_not_bound` and sends nothing. It fails closed. Live formatting is criterion 10, which this PR leaves for Fisher.

PR #110, opened 2026-08-28
feat(desktop): export a redacted diagnostic bundle for one failed run

### LINKED ISSUE #75: P1: Export a redacted diagnostic bundle for one failed run

FisherXZ:
Closes #75.

## Domain words

- **Bundle** — one local ZIP describing one failed run or one state finding.
- **Subject** — the exact run id, or the exact Doctor finding, an export started from. Every bundle names one.
- **Member** — one file inside the archive.
- **Scan** — the pre-export check that refuses to build a bundle at all.
- **Fail closed** — a matched scan produces nothing. Not a redacted bundle, not a warning file, not an empty folder.

## Why this is a security change, not a diagnostics change

A diagnostic bundle aggregates everything about a failed run into one file whose entire purpose is to be handed to someone else. Every other safety property in Sourcecado is undone if a secret rides out in here.

So the deliverable is the redaction layering and the fail-closed path. The packaging is the easy part.

## Three independent redaction layers

| Layer | Source | Independent because |
|---|---|---|
| 1 | The run receipt's closed field allowlist, from #74 | Already merged; the bundle consumes receipts rather than raw checkpoints |
| 2 | Doctor's bounded redacted output contract, from #73 | Already merged; reused rather than reimplemented |
| 3 | The bundle's own pre-export scan | **Imports neither of the first two** |

The third layer's independence is the point. A scan that trusts the layers above it keeps passing after one of them changes — which is how a leak appears months later, in a build nobody thought they had altered.

Two tests hold that line:

- `test_the_scan_still_catches_a_secret_when_the_upstream_layers_are_disabled` — the scan is exercised with layers 1 and 2 turned off, and still catches the secret.
- `test_the_third_layer_imports_neither_of_the_first_two` — structural, so the independence cannot be quietly undone by an import.

## Fail closed means produce nothing

A matched scan aborts the export. There is no redacted-and-continue path.

The reasoning: a scan that matched is evidence that something reached the bundle by a route nobody modeled. Producing a "cleaned" bundle from a state you have just proved you do not understand is worse than producing nothing.

`test_a_scan_failure_names_a_category_and_a_location_but_never_the_value` — the operator is told what category and where, never the matched value. An error message is not a safe place to print a secret.

Both pattern-based and registered-secret scans run. `test_a_registered_secret_reaching_the_bundle_fails_the_export_closed` and `test_a_credential_pattern_reaching_the_bundle_fails_the_export_closed`.

## No partial artifact, ever

Assembly happens in a temporary location and moves into place only after every scan passes.

- `test_a_failed_scan_leaves_no_file_and_no_temporary`
- `test_a_crash_while_assembling_leaves_no_file_and_no_temporary` — forces a mid-build exception and asserts nothing readable remains anywhere.

A successful export writes one file at mode 0600: `test_a_successful_export_writes_one_owner_only_file`.

## Planted secrets, proven non-vacuously

Every category the issue names is planted: logs, errors, connector payloads, paths, and approval data.

The non-vacuity guard is a separate test that runs first: `test_every_planted_category_is_carried_by_the_bundle_before_any_leak_check` asserts the bundle actually carries evidence for each category. Only then does `test_no_planted_secret_survives_into_any_member` assert absence. Without the first, the second would pass when a category produced no output at all — and would keep passing after the redaction broke.

Raw home paths are covered specifically, since `/Users/<name>/…` in a traceback identifies a real person: `test_a_home_path_in_carried_text_becomes_the_home_placeholder` and `test_a_raw_home_path_never_survives`.

`test_prompts_and_message_bodies_have_no_field_to_land_in` — the exclusion is structural, not a filter.

## Reused vocabulary rather than a parallel one

The bundle expresses "we do not know" using `run_evidence.py`'s existing closed vocabulary — `present`, `absent`, `partial`, `missing`, `ambiguous`, `unsupported`, `expired` — rather than coining its own. `test_the_bundle_reuses_the_merged_evidence_vocabulary` and `test_absent_and_missing_stay_different_through_the_projection`.

Two vocabularies for the same distinction drift apart, and this one had already been thought through in #74.

## Never uploads

`coworker/diagnostic_bundle.py` opens no network client and imports nothing that could. `test_the_bundle_module_opens_no_network_client` holds it.

The operator surface is Settings → Diagnostic bundle, deliberately two separate steps: **Review bundle** assembles, scans, and shows a bounded preview with every category and what is left out, writing nothing; **Save bundle to this Mac** repeats the work and writes one file, reporting path, size, and SHA-256.

## Determinism

`test_two_exports_from_identical_state_differ_only_in_package_identity`, and `test_the_document_builder_has_no_clock_and_no_randomness` — the builder cannot introduce nondeterminism because it has no access to either.

## Measurements

```
1107 passed, 3 skipped, 1 warning in 43.48s
 Test Files  46 passed (46)
      Tests  419 passed (419)
```

All re-run independently by the reviewing session on the pushed commit. 35 tests in `test_diagnostic_bundle.py`.

## Boundary

Reads Doctor and ledger output; modifies neither. No file under `agent_run*`, `run_ledger`/`run_receipt`/`run_evidence`, `migrations.py`, `doctor.py`, `turn.py`, `store.py`, `people.py`, `permissions.py`, or `provider.py` was touched.

### DIFF
--- a/desktop/coworker/bundle_redaction.py
+++ b/desktop/coworker/bundle_redaction.py
@@ -1,0 +1,272 @@
+"""The last line of defence before a diagnostic bundle leaves this machine.
+
+Two upstream layers already bound what a bundle can carry: the run receipt's
+closed field allowlist, and the state report's own bounded, redacted output
+contract. This module is the third, and it is deliberately independent of both.
+
+Independence is the whole point. A scan that imported an upstream redactor
+would keep passing after that redactor changed, which is exactly the failure a
+third layer exists to catch. So this file imports nothing from the rest of
+Sourcecado, carries its own pattern vocabulary, and is tested on its own.
+
+Two jobs:
+
+`relativize` and `scrub` rewrite absolute paths before anything is packaged.
+A path anchored in the operator's home directory identifies a real person, so
+the whole path token collapses to `<home>`; a path inside the state directory
+keeps its remainder, because `<state>/club.db` is diagnostic and safe.
+
+`scan` refuses. It reports a category and a location and never the matched
+value, so a refusal can be logged and shown without becoming the leak it was
+trying to prevent.
+"""
+
+from __future__ import annotations
+
+import re
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Iterable
+
+STATE_LABEL = "<state>"
+HOME_LABEL = "<home>"
+
+# The shortest string in the registered store worth treating as a credential.
+# Below this a value is a connector name, a scope word, or a status.
+MIN_REGISTERED_LENGTH = 12
+
+
+@dataclass(frozen=True)
+class ScanMatch:
+    """One refusal. Carries where and what kind, never the value itself."""
+
+    category: str
+    location: str
+
+
+SCAN_CATEGORIES = frozenset(
+    {
+        "registered_secret",
+        "private_key",
+        "json_web_token",
+        "issued_credential",
+        "authorization_header",
+        "credential_assignment",
+        "home_path",
+    }
+)
+
+_PRIVATE_KEY = re.compile(
+    r"-----BEGIN (?:[A-Z0-9 ]+ )?PRIVATE KEY-----", re.IGNORECASE
+)
+_JSON_WEB_TOKEN = re.compile(
+    r"\beyJ[A-Za-z0-9_-]{8,}\.[A-Za-z0-9_-]{8,}\.[A-Za-z0-9_-]{8,}"
+)
+# Issuer prefixes that mean "credential" on their own. Each one is a published
+# token format, so a match here is never a guess about entropy.
+_ISSUED_CREDENTIAL = re.compile(
+    r"(?:sk-(?:ant-|proj-|live-|test-)?[A-Za-z0-9_-]{16,}"
+    r"|rk-[A-Za-z0-9_-]{16,}"
+    r"|pk_(?:live|test)_[A-Za-z0-9]{16,}"
+    r"|gh[pousr]_[A-Za-z0-9]{16,}"
+    r"|github_pat_[A-Za-z0-9_]{20,}"
+    r"|glpat-[A-Za-z0-9_-]{16,}"
+    r"|xox[baprs]-[A-Za-z0-9-]{16,}"
+    r"|shpat_[A-Za-z0-9]{16,}"
+    r"|dop_v1_[A-Za-z0-9]{16,}"
+    r"|npm_[A-Za-z0-9]{20,}"
+    r"|SG\.[A-Za-z0-9_-]{16,}\.[A-Za-z0-9_-]{16,}"
+    r"|A(?:KIA|SIA)[0-9A-Z]{16}"
+    r"|AIza[A-Za-z0-9_-]{20,}"
+    r"|ya29\.[A-Za-z0-9._-]{12,}"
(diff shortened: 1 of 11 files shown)


FisherXZ (commented):
## Verdict
**Ship.**

## Summary
An operator can review and save one local diagnostic ZIP for a failed run. The zip joins run evidence, versions, health, state, connectors, and structured logs. Prompts, bodies, tool payloads, and credentials have no field to land in. Free text from a run is a length.

Three redaction layers stay independent. The third layer scans the projected document and the member bytes, then refuses. A match writes nothing. Assembly stages a 0600 file and only `os.replace` puts it in place.

The GUI is Review, then Save bundle to this Mac. The sidecar never uploads. Offline inspect verifies the manifest and checksums.

## Map
```mermaid
flowchart TD
  A[Review bundle] --> B[Resolve run or Doctor finding]
  B --> C[Read receipt, doctor, connectors, events]
  C --> D[Project onto closed field lists]
  D --> E[Rewrite home paths]
  E --> F[Scan document]
  F -->|match| G[Refuse: category and location only]
  F -->|clean| H[Bounded preview. Nothing written]
  H --> I[Save bundle to this Mac]
  I --> J[Rebuild. Scan document. Scan member bytes]
  J -->|match| G
  J -->|clean| K[Stage owner-only zip in private temp]
  K --> L[Atomic replace into diagnostics/]
  L --> M[Show path, size, SHA-256]
  G -.-> N[No file. No temp. No empty folder]
```

Why Ship: #75 is a security export, not a log dump. The design keeps the third scan independent of the receipt allowlist and of Doctor, fails closed, and leaves no partial artifact. I loaded #75, `docs/diagnostic-bundle.md`, and the merge-base diff against `7563a77`. I ran `cd desktop && .venv/bin/pytest -q` on `2e3b3a8`: 1107 passed, 3 skipped. GUI: 46 files, 419 tests passed. A planted failed-run export had no canary in any member.

## Blocking
None.

The failure I was most afraid of was a secret from a failed send riding out in the zip, or a refused export leaving that zip on disk. I exported a planted failed run on this head. The canaries were gone. The scan did not match. A rename crash left nothing readable.

PR #111, opened 2026-08-28
feat(desktop): detect stale run owners and impossible run history in Doctor

FisherXZ:
Completes #73. PR #103 landed the migration registry, integrity checks, backup, repair, and the redacted report, and deliberately deferred three checks that needed a canonical Agent Run identity. #105 merged that identity. This closes them.

## Domain words

- **Owner** — the process currently driving an Agent Run. Holds an exclusive `flock` on a marker file for its whole life.
- **Lease** — an owner's time-bounded claim on a run.
- **Liveness** — the three-state answer to "is that owner still alive": `ALIVE`, `DEAD`, `UNKNOWN`.
- **Kernel proof** — the only acceptable evidence that an owner is dead: the kernel released its `flock`, which happens on process exit by any means and never before.

## The three deferred checks

1. **Stale run owner and lease detection** — `agent_run.dead_owner`, `agent_run.expired_lease`, `agent_run.unknown_owner`, `agent_run.owned_by_live_process`, `agent_run.unowned_running`.
2. **Impossible checkpoint transition detection.**
3. **Orphaned queue and approval links** keyed by canonical run identity.

## The rule that governs all of it

**Doctor may report an owner as dead only on kernel proof. Never on a heuristic, never on a timestamp, never on a process id missing from `ps`.**

Collapsing `UNKNOWN` into `DEAD` is the failure that matters here. It would let Doctor recommend reclaiming a run whose owner is alive and working — which is precisely what the `flock` design in #105 exists to prevent. Two writers on one run is exactly the corruption the whole ownership model was built to make impossible.

An unknown owner does not need rescuing. Its lease still expires, and an expired lease is version-fenced, so reclaiming it later still cannot produce two writers. Doctor's job here is to report, not to rescue.

The distinction is carried into the operator-facing text, not just the code: the unknown finding's summary contains **"Unknown is not dead"**.

## The tests that hold it

These use real processes and real kernel behavior, not mocks.

`test_a_live_run_owner_is_never_reported_as_stale` spawns a child that holds an exclusive `flock` on its owner marker, runs Doctor while it is genuinely alive, and asserts:

- the run is reported as `owned_by_live_process`, severity `INFO`, repair `NONE`
- `dead_owner`, `expired_lease`, and `unowned_running` are all **absent**
- `report.healthy is True` — a live owner is not a defect
- no repair is proposed against `agent_runs_db`

`test_a_dead_owner_is_reported_only_on_kernel_proof` spawns a child that exits, so the kernel releases the lock, then asserts the dead finding appears with repair `REVIEW_REQUIRED`.

Three separate unknown cases, each asserting `dead_owner` is absent:

| Test | Situation |
|---|---|
| `test_a_missing_owner_marker_is_unknown_and_never_dead` | Marker file gone |
| `test_an_owner_on_another_host_is_unknown_and_never_dead` | Different host |
| `test_a_run_store_with_no_owner_directory_reports_unknown_not_dead` | No owner directory at all |

And `test_an_expired_lease_is_reported_apart_from_a_dead_owner` keeps those two situations distinct, because they call for different operator responses.

## Repairs

Every new finding is report-only or `REVIEW_REQUIRED`. None is automatic.

Run ownership is the domain where an "obvious" automatic repair does real damage, and no repair here is provably safe against a live owner. Doctor stays diagnostic, which is the contract PR #103 established.

## Measurements

```
1070 passed, 3 skipped, 1 warning in 43.14s
```

## Boundary

Three files: `doctor.py`, `test_doctor.py`, `doctor.md`. The `agent_run*` modules are read, never modified — they are merged, reviewed correctness from #105 and #63.

With this, #73 is complete as originally specified.

### DIFF
--- a/desktop/coworker/doctor.py
+++ b/desktop/coworker/doctor.py
@@ -10,6 +10,12 @@ left at the end of an append-only log, and applying a registered migration. A
 corrupt row, an orphaned link, an interrupted approval, or anything touching a
 person file or an external outcome is reported for review and left alone.
 
+Agent Run ownership is report-only, without exception, and one rule governs it:
+an owner is called dead only on kernel-backed proof that its process has exited.
+Another host, a missing marker, or a platform without file locking are unknown,
+and unknown is reported as unknown. Rounding unknown up to dead would invite an
+operator to take a run away from a process that is still working it.
+
 Output is bounded and redacted. Doctor reports store ids, versions, check names,
 and counts. It never prints a row, a message body, a token, an authorization
 header, or a path outside the state directory.
@@ -31,15 +37,28 @@ from pathlib import Path
 from typing import Any
 
 from coworker import migrations
+from coworker.agent_run_owner import OWNERS_DIR_NAME, Liveness, OwnerRegistry
+from coworker.agent_run_repository import DB_NAME as AGENT_RUNS_DB
+from coworker.agent_run_repository import SCHEMA_VERSION as AGENT_RUNS_VERSION
+from coworker.agent_run_state import (
+    AgentRunTransitionError,
+    is_leasable,
+    is_terminal,
+    validate_transition,
+)
+from coworker.agent_runs import json_list
 from coworker.migrations import (
     BackupFailed,
     StoreKind,
     StoreStatus,
+    VersionChannel,
     open_readonly,
     state_root,
     store_path,
     store_present,
 )
+from coworker.run_evidence import analyze_record
+from coworker.run_receipt import parse_moment
 
 MAX_DETAIL_ROWS = 8
 MAX_DETAIL_CHARS = 200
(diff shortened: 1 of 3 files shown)


FisherXZ (commented):
## Verdict
**Ship.**

## Summary
Doctor now inspects Agent Run owners, history, and run-keyed links. An owner is dead only when `OwnerRegistry.liveness_of` takes a free exclusive `flock`. A missing marker, another host, a missing marker directory, and a loosened marker directory stay unknown. Expiry is decided before liveness, so a live process holding an expired lease is an expired lease, not a dead owner. History walks stored checkpoints through `validate_transition` and `analyze_record`. Pruned tails are not gaps. Every Agent Run finding is report-only.

This closes the three checks issue #73 deferred after #103, once #105 landed the canonical run identity.

## Map
```mermaid
flowchart TD
  diagnose[Doctor diagnose] --> leases{Unexpired lease?}
  leases -->|no| expired[expired_lease]
  leases -->|yes| probe{Marker dir owner-only?}
  probe -->|no| unknown[unknown_owner]
  probe -->|yes| flock[OwnerRegistry.liveness_of]
  flock -->|lock held| alive[owned_by_live_process]
  flock -->|lock free| dead[dead_owner]
  flock -->|other host or missing marker| unknown
  diagnose --> history[validate_transition]
  diagnose --> report[Report only]
  alive --> report
  dead --> report
  unknown --> report
  expired --> report
  history --> report
```

Why Ship: The #105 rule holds in Doctor. Dead requires kernel proof. Unknown stays unknown. Doctor writes nothing to a run store. I reran `cd desktop && .venv/bin/pytest -q` on `a5effc2`: 1070 passed, 3 skipped, 1 warning in 42.68s.

## Blocking
None.

The failure I was most afraid of was Doctor calling a live or unprovable owner dead. `test_a_live_run_owner_is_never_reported_as_stale` holds a real `flock` in a child process and still reports the install healthy. Collapsing unknown into dead on this head failed the four unknown-owner tests.

Issue #65, opened 2026-08-27
P1: Treat connector and tool output as untrusted evidence end to end


FisherXZ:
## What to build

Introduce a Sourcecado-owned untrusted-content boundary for Gmail, Drive, Calendar, Granola, web, MCP, workspace, and shell results. External content may supply evidence but cannot redefine product policy, mint authority, choose an approval outcome, broaden tool access, or silently convert itself into durable truth.

The taint decision must survive provider retries, context assembly, compaction, Agent Run checkpoints, and restart recovery.

## Acceptance criteria

- [ ] Every external result carries a trusted origin classification and stable source reference before entering model context.
- [ ] Prompt-facing external text is clearly delimited as untrusted evidence, with forged delimiters escaped.
- [ ] External instructions cannot override enrichment, send, filesystem, shell, memory, or Board policy.
- [ ] Tainted content cannot create standing authority or alter approval scope.
- [ ] Automatically allowed record writes still require Sourcecado-owned schemas, actor/run identity, source references, and rationale.
- [ ] Taint remains sticky across retries, compaction, provider failover, semantic checkpoints, and restart resume.
- [ ] Logs, diagnostics, approvals, and UI receipts retain origin without reproducing unsafe bodies.
- [ ] Malicious fixtures cover Gmail, Drive documents, Granola notes, web snippets, MCP results, and shell output.
- [ ] Tests prove the same content typed directly by the director and received from an external source are not granted the same authority.

## Blocked by

- #54
- #55
- #84
## Sprint 5 Evidence Envelope contract

Authoritative plan: [Sourcecado Sourcing OS Release Program — Execution Plan](https://app.notion.com/p/3c9d64acf06681af91dbee2d65715f70)

Implement one Sourcecado Evidence Envelope containing stable source identity, origin, sensitivity, freshness, truncation, trust classification, and visible source reference. Adapt Gmail, Drive, Calendar, Granola, Apollo, web, MCP, filesystem, and shell output into this contract before the content enters model context or durable domain projections.

### Additional release acceptance criteria

- [ ] Forged evidence delimiters are escaped before model context.
- [ ] Taint remains sticky through retry, failover, compaction, checkpoints, person projection, and restart.
- [ ] The same text typed by the director and received from an external source retains different authority.
- [ ] The envelope is the shared input to person evidence, the living brief, run receipts, and diagnostic redaction.

## Reference codebases

Implementation may consult these local repositories as read-only evidence when relevant:
[...]

PR #114, opened 2026-08-28
feat(desktop): treat connector output as untrusted evidence end to end

FisherXZ:
Addresses #65. This is the Evidence Envelope and its adapters; #84's certification is separate.

## Domain words

- **Envelope** — one record holding text somebody other than Sourcecado wrote, plus who wrote it.
- **Directive** — a record carrying director intent. Only the director channel can mint one.
- **Origin** — who wrote the bytes, not who fetched them: `director`, `sourcecado`, or `external`.
- **Fence** — the delimited block external text appears in when it reaches model context.

## Problem

Connector output was ordinary text by the time it reached the model. A Gmail body saying "enrich this candidate and email them" was, structurally, indistinguishable from the director typing it.

## Criterion 9 is solved by types, not by filtering

The requirement is that the same sentence carry different authority depending on who wrote it. That is not a filter you can add — it is an authority model.

`Trust` and `Authority` are derived from `Origin` by a total map and are **never constructor arguments**. `Envelope` is frozen and assigns both in `__post_init__`. There is no keyword, no payload key, and no connector field that yields trusted external content. A Granola note shipping `"trust": "authoritative"` at its top level is bytes inside `body`.

Only the director channel mints a `Directive`, and only a `Directive` carries `director_intent`. **An `Envelope` has no path to that type.** The text is identical; the records are not.

`test_an_approval_quoting_the_mail_is_marked_and_clamped` and `test_an_approval_the_director_asked_for_keeps_its_scope` are that pair.

## Criterion 2: the fence cannot be closed by its own content

Two independent defenses, not one.

**Line anchoring.** Every body line is prefixed with `| `, and the fence is line-anchored. Content cannot produce an unprefixed line, so it cannot produce a fence, whatever it contains and however it spaces it. Bodies split with `str.splitlines`, which honours `\v`, `\f`, `\x1c`–`\x1e`, `\x85` and the Unicode separators — an exotic separator becomes another prefixed line rather than a way out.

**A nonce.** 128 fresh random bits per block, verified absent from the body. `seal` checks its own output with `fence_intact` and **raises rather than shipping a block it cannot verify**, so an escaping bug fails the call instead of quietly opening a hole.

Runs of three or more `<` or `>` are broken with hyphens — deliberately lossy, for readers that strip prefixes, and so an operator can see the attempt.

Fixtures try all of it: `FORGED_OPEN`, `FORGED_CLOSE`, `FORGED_CLOSE_SPACED`, and `FORGED_CLOSE_NEAR_MISS`. `test_the_forged_fence_in_the_mail_does_not_split_the_block`.

## Criterion 6: origin survives in the id itself

A reference reads `ext_gmail_<16 hex>`, `own_board_<16 hex>`, or `dir_chat_<16 hex>`, and `origin_of_ref` recovers the origin **from the id alone**. A transport that keeps only ids still knows the content was external.

That matters because `agent_runs.CHECKPOINT_PAYLOAD_FIELDS` keeps `source_ref_ids` and drops everything else — no origin field, no trust field, no body. Encoding origin in the id is what makes taint survive a layer that was designed to throw structure away.

Anything `origin_of_ref` cannot parse is external. Every pre-existing reference shape in this codebase — `drive:<file>:<mtime>`, `meeting_<digest>`, a raw Drive file id — is connector-supplied, so failing closed is correct as well as safe.

One test per transition:

| Transition | Test |
|---|---|
| Provider retry rebuilding the request | `test_taint_survives_a_provider_retry_that_rebuilds_the_request` |
| Provider failover to another model | `test_taint_survives_provider_failover_to_another_model` |
| Transcript compaction | `test_taint_survives_transcript_compaction` |
| Semantic checkpoint | `test_taint_survives_a_semantic_checkpoint` |
| Restart reloading the transcript | `test_taint_survives_a_restart_that_reloads_the_transcript` |
| Resumed turn's next read | `test_a_resumed_turn_still_fences_the_next_read` |

## Criterion 3: hostile content changes no policy

| Test | Proves |
|---|---|
| `test_a_mail_demanding_a_send_still_stops_at_the_human_gate` | Injected send instruction does not send |
| `test_no_hostile_source_changes_a_permission_class` | No source moves a tool between `ASK`/`AUTO`/`RETRY_SAFE` |
| `test_a_mail_cannot_add_a_tool_to_the_run` | External content cannot broaden the tool catalog |
| `test_the_tool_finished_event_names_the_source_without_quoting_it` | Criterion 7 — receipts keep origin without reproducing unsafe bodies |

## Six malicious fixtures, one per source

`gmail_message`, `drive_document`, `granola_note`, `web_snippet`, `mcp_result`, `shell_output` — written separately rather than one fixture parameterized six ways, because the injection surface differs per connector. Each carries instruction injection, forged delimiters, and credential-shaped strings. Credential fixtures are assembled at runtime, so no credential-shaped literal sits in source.

## Non-vacuity

`test_deleting_the_fail_closed_default_would_be_caught` removes the fail-closed default and asserts the suite goes red. Without it, the default could stop working silently and every other test would still pass.

## Measurements

```
1350 passed, 3 skipped, 1 warning in 63.49s
 Test Files  49 passed (49)
      Tests  441 passed (441)
```

Re-run independently by the reviewing session on the pushed commit, rebased onto `e91f1d0`.

## Rebase note

One conflict, `apollo.py`, against #109's `enrichment_match` and `enrichment_resource`. Purely additive on both sides — main added the enrichment approval helpers, this branch added `apollo_evidence`. Both kept. The conflict ran to end of file with nothing following it, so the hoisted-shared-line hazard a sibling lane hit did not apply here; verified by AST parse and by the full suite.

## Boundary

`store.py`, `scheduler.py`, `permissions.py`, `people.py`, `migrations.py`, `doctor.py`, `run_ledger*`, `context_projection.py`, the `agent_run_*` modules, and the prompt/persona set were not touched. `permissions.py` in particular is read-only here — criterion 3 is tested against the real `ASK`/`AUTO`/`RETRY_SAFE` sets rather than a mock.


FisherXZ (commented):
## Verdict
**Fix, then ship.**

## Summary
This PR adds one Evidence Envelope for connector output. Trust is derived from origin. The fence is line-anchored and nonce-bearing. Adapters cover Gmail, Drive, Calendar, Apollo, web, MCP, filesystem, and shell. The same sentence from the director and from Gmail becomes two records.

The turn still keeps a full workspace body in memory after `fs_read` or `shell_exec`. `_persist_tool_result` redacts before it writes, so a clean finish stays redacted. `_persist_closed` then rewrites the transcript from that in-memory message. `sanitize_message` still looks for `content` and `output`. Those keys now live inside `evidence`. Cancelling a later approval writes the file body into the conversation database.

## Map
```mermaid
flowchart LR
  read["fs_read / shell_exec"] --> mem["In-memory fenced body"]
  read --> persist["_persist_tool_result"]
  persist --> disk["Disk redacted"]
  mem --> later["Later write approval cancelled"]
  later --> closed["_persist_closed"]
  closed -.-> leak["Disk keeps the workspace body"]
```

Why Fix, then ship: Criterion 2 and the authority model hold on the Gmail path. I reran `cd desktop && .venv/bin/pytest -q` on cdf9359: 1350 passed, 3 skipped. The workspace two-tier redaction this PR claims to preserve does not hold on the cancel path.

## Blocking
1. After Sourcecado reads a workspace file in a turn, cancelling a later write approval stores that file body in the conversation database. `sanitize_message` still strips `content` and `output`. Fencing moved the body into `evidence`. `_persist_closed` then replace_alls the unredacted in-memory tool message. I reproduced this on cdf9359: `disk_has_secret True` after `wait_permission` returned `cancel`.

FisherXZ (commented):
## Verdict
**Ship.**

## Summary
The Evidence Envelope still treats connector output as untrusted. Trust is derived from origin. The fence stays line-anchored.

Round one blocked the cancel path: after an `fs_read`, cancelling a later write approval wrote the workspace file body into the conversation store. `a6406c0` strips the fenced `evidence` body in persist-time sanitization, the same layer `_persist_closed` already uses.

## Map
```mermaid
flowchart LR
  read["fs_read"] --> mem["In-memory fenced body"]
  read --> persist["_persist_tool_result"]
  persist --> disk["Disk redacted"]
  mem --> later["Later write approval cancelled"]
  later --> closed["_persist_closed"]
  closed --> sanitize["sanitize_message strips evidence"]
  sanitize --> disk
```

Why Ship: the round-one cancel leak is closed. I reran `test_cancelling_a_later_approval_does_not_write_a_workspace_file_body_to_disk` on cdf9359: it failed with the body in `store.load`. On a6406c0 it passed. Desktop pytest: 1351 passed, 3 skipped.

## Blocking
None.

The failure I was most afraid of was the cancel path still replace_alling the unredacted in-memory tool message. That path is covered on this head and it passed.

PR #116, opened 2026-08-28
feat(desktop): verify legal artifacts before offering them as usable

FisherXZ:
Addresses the code portion of #39. **It does not close the issue** — see "What this does not do".

## Domain words

- **Facet** — one of the four things that must verify before a legal artifact can be offered: parties, dates, terms, approval.
- **Declared status** — what the caller says an artifact is. Never inferred from the body.
- **Ready to use** — the only state in which Sourcecado will present an artifact as usable.

## Problem

A stale NDA exists whose body names De Beers and Berkeley Consulting — parties that do not match its title or intended use. Sourcecado had one narrow hardcoded guard against exactly that document, and no general rule that legal artifacts are evidence rather than agreements.

## Where the stale NDA actually is

**Not in this repository.** It is a live Google Drive file, `Codeology NDA Template.docx`, referenced by a test fixture at `desktop/tests/test_drive.py:701-724` and by the QA finding at `docs/qa/2026-08-26-pr35-full-user-trail.md:65`. There is nothing here to delete or rename.

So "quarantine" in code means: a document of this shape can never be classified `approved_template` or `ready_to_use`, whatever it is titled. Renaming or moving the actual Drive file remains a human action, already recorded in `docs/runbooks/p0-source-safety-remediation.md` section 4.

## The state model

`draft`, `approved_template`, `executed`, `stale`, `unverified`.

Status is **always caller-declared and never inferred from body wording**. Inferring it would let a well-worded fake earn `approved_template` by containing the right phrases — precisely the attack this exists to stop.

`resolve_status()` maps any undeclared or unrecognized value to `unverified`, **never** to `draft`. An unplaced document is unsafe, not "in progress".

`ready_to_use` requires `status == approved_template` **and** all four facets `present`. Every other status is hard-gated to false even when all four facets read perfectly clean:

- `test_stale_status_is_never_offered_as_ready_regardless_of_facets`
- `test_undeclared_status_blocks_readiness_even_with_clean_facets`

That gate is criterion 2 — `stale` and `unverified` cannot be confused with `approved_template` — enforced structurally rather than by convention.

## The four facets

| Facet | Verified from | Non-present outcomes |
|---|---|---|
| parties | The body's own "between X and Y" language, **never the title** | `absent` if the expected party is not named; `partial` if an extra non-placeholder party appears; `missing` if there is no party language at all |
| dates | A real date or an explicit placeholder in the body | `missing` |
| terms | A term/duration marker or placeholder | `missing` |
| approval | A caller-supplied record | `missing` with no record; `absent` when `authorized` is false; **`expired` when `approved_at` precedes the artifact's `modified_time`** |

That last one is the interesting case: the body changed after approval, so the approval no longer describes the document it approved.

Reading parties from the body rather than the title is the whole point — the headline failure is a filename that says one thing while the body names someone else.

Any non-`present` facet blocks readiness and produces an inspectable reason, e.g. `parties:expected_party_not_named:...` or `approval:approval_superseded_by_later_revision`.

## Tests

The four cases criterion 5 names, each non-vacuous — the facet that should be clean is asserted clean **first**, proving the body was really parsed and that a different check is not what tripped it:

| Case | Test |
|---|---|
| Filename/body party mismatch | `test_classify_flags_filename_body_party_mismatch` |
| Missing dates | `test_classify_flags_missing_dates` |
| Stale approval | `test_classify_flags_stale_approval` |
| A valid reviewed template | `test_classify_verifies_a_valid_reviewed_template` |

The fourth matters as much as the others: a checker that refuses everything is useless. It asserts full dict equality on all four facets, `ready_to_use is True`, and `knowledge_gap_fields() is None`.

16 tests total. All fixtures use fictional parties (Northwind Distribution, Ridgeline Ventures). No real party name, agreement text, or secret was introduced.

## Measurements

```
baseline  origin/main e91f1d0 : 1280 passed, 3 skipped
after                          : 1296 passed, 3 skipped
```

Baseline measured by running it, not by arithmetic. Delta is exactly the 16 new tests; nothing else moved. Re-run independently by the reviewing session on the pushed commit.

## What this does not do

Stated plainly because it matters more than what it does.

**Nothing calls `legal_artifacts.classify()` yet.** The module is a correct, tested engine that is not in a live path. The only thing currently guarding a real Drive read is the pre-existing hardcoded hotfix at `drive.py:89-100`, which matches literally `"codeology"` in the name and `"berkeley consulting"` in the body. This PR generalizes that logic; it does not yet replace it.

Wiring it means touching `drive.py`, which **open PR #114 also modifies**. Doing it here would have created a conflict for a change that is better reviewed on its own. That is a deliberate deferral, not an oversight.

Three things remain for a human, and #39 should stay open until they are done:

1. An authorized Codeology officer or counsel authors the canonical Berkeley Codeology template text and records approval (`approved_by`, `approved_at`, `authorized: true`) plus the canonical Drive id and version. **This PR defines the placeholder structure and the checks a template must pass — not the legal language.**
2. An authorized Drive owner renames or moves the live `Codeology NDA Template.docx` so it is visibly quarantined in Drive itself. Code cannot do this.
3. Wiring `classify()` into the Drive read path, after #114 merges.

## Open question for review

"Terms" verification is structural — it looks for a term or duration marker or placeholder, not for clause semantics. That was deliberate, to avoid the module making legal judgments. If the product wants richer clause-presence checking (confidentiality scope, governing law), that is a product decision, not something to infer.

## Summary by CodeRabbit

- **New Features**
  - Added legal document classification to distinguish verified, incomplete, stale, and unverified artifacts.
  - Documents are marked ready for use only when approved and all required details are confirmed.
  - Added detection of missing information, placeholders, outdated approvals, and knowledge gaps.
  - Added support for recording knowledge gaps with duplicate prevention.

- **Documentation**
  - Documented classification rules, readiness requirements, and knowledge-gap behavior.

- **Tests**
  - Added comprehensive coverage for verification, approval validation, readiness checks, and gap handling.


FisherXZ (commented):
## Verdict
**Fix, then ship.**

## Summary
This PR adds `coworker.legal_artifacts.classify()`. It reads parties, dates, and terms from the body, not the filename. Status is caller-declared. `ready_to_use` is true only for `approved_template` with all four facets present. Sixteen new tests. Measured `1296 passed, 3 skipped` on this head. Nothing calls `classify()` yet. Drive still uses the hardcoded Codeology / Berkeley Consulting hotfix.

The approval facet does not parse timestamps and treats any truthy `authorized` as a real review. A clean body with `approved_at="not-a-timestamp"`, with no `modified_time`, or with `authorized="false"` still returns `ready_to_use: True`.

## Map
```mermaid
flowchart TD
  A[Caller classify] --> B[Resolve declared status]
  B --> C[Read parties from body]
  C --> D[Read dates and terms from body]
  D --> E[Check approval record]
  E --> F{authorized truthy and approved_at nonempty?}
  F -->|no| G[approval missing or absent]
  F -->|yes| H{modified_time string greater than approved_at?}
  H -->|yes| I[approval expired]
  H -->|no or modified_time missing| J[approval present]
  J -.-> K["ready_to_use True if approved_template"]
  G --> L[ready_to_use False]
  I --> L
```

The dashed edge is the miss: an unparseable `approved_at` or a missing revision time still lands on `approval present`.

Why Fix, then ship: Issue #39 says Sourcecado must not label a legal artifact ready without verifying approval. The engine is the right seam. Parties, dates, terms, and the status gate fail closed. The approval check fail-opens. That is the criterion the stale-approval test exists to protect. Do not merge until that facet fail-closes.

## Blocking
1. `classify()` labels a reviewed-looking template ready when the approval record is not a real, comparable review. I ran the valid-template fixture from `test_classify_verifies_a_valid_reviewed_template`. `approved_at="not-a-timestamp"` returned `ready_to_use True` and `approval.evidence=present`. Omitting `modified_time` did the same. `authorized="false"` did the same. Boolean `False` correctly blocked. `_verify_approval` only checks nonempty strings and truthiness, then compares them lexicographically. Issue #39 requires verifying approval status before any ready-to-use label.

FisherXZ (commented):
## Verdict
**Ship.**

## Summary
The round-one blocker is closed. `d14fdfd` makes `classify()` require boolean `authorized=True` and ISO-8601 `approved_at` plus `modified_time` before approval can be `present`. The three cases that returned `ready_to_use True` on `1ff637d` now return False. Nineteen tests. Measured `1299 passed, 3 skipped` on this head. Still nothing calls `classify()`. Drive still uses the hardcoded Codeology / Berkeley Consulting hotfix.

## Map
```mermaid
flowchart TD
  A[Caller classify] --> B[Resolve declared status]
  B --> C[Read parties dates terms from body]
  C --> D[Check approval record]
  D --> E{authorized is boolean True?}
  E -->|no| F[approval absent]
  E -->|yes| G{approved_at and modified_time parse as ISO-8601?}
  G -->|no| H[approval missing]
  G -->|yes| I{modified after approved_at?}
  I -->|yes| J[approval expired]
  I -->|no| K[approval present]
  K --> L{approved_template and all facets present?}
  L -->|yes| M[ready_to_use True]
  L -->|no| N[ready_to_use False]
  F --> N
  H --> N
  J --> N
```

The parse check is now a solid reject. The round-one dashed miss is gone.

Why Ship: Issue #39 requires verifying approval before any ready-to-use label. I reran the three round-one cases on `d14fdfd`. Unparseable `approved_at`, omitted `modified_time`, and `authorized="false"` all fail closed. The valid-template fixture still returns ready.

## Blocking
None.

The failure I was most afraid of was `approved_at="not-a-timestamp"` still labeling the valid-template fixture ready. On this head that call returns `ready_to_use False` and `approval.evidence=missing`.

PR #113, opened 2026-08-28
feat(desktop): file inbound Gmail replies and show follow-up on the Board

FisherXZ:
Reopened against `main` after #109 merged and GitHub closed the stacked PR (#112) because its base branch was deleted.

Closes #67.

> **Stacked PR.** Base is `claude/s3-approved-send` (#109), not `main`, so the diff shows only this issue's work. Retarget to `main` once #109 merges. #109 satisfies this issue's first acceptance criterion directly.

## Domain words

- **Tracked thread** — a Gmail thread Sourcecado knows about because it sent into it and stored the `message_id`/`thread_id`.
- **Filing** — attaching an inbound reply to exactly one person.
- **Unassigned** — a reply we can see but will not attribute. It becomes a reviewable knowledge gap, not a guess.
- **Cursor** — the durable Gmail history boundary that makes refresh incremental instead of a full rescan.

## Problem

Sourcecado sent mail and then stopped watching. A reply arriving was invisible, the Board could not show who was waiting on a follow-up, and nothing moved a person from Open to In conversation on real evidence.

## The design decision that matters: refusing to guess

The natural implementation is to file a reply on the best-matching person. That is wrong, and it is wrong in a way that is expensive: a reply filed on the wrong person contaminates that person's evidence and can drive a bad outreach decision, and nothing downstream would know.

So this files only on exact thread and recipient evidence, and refuses in every case where the evidence is not exact. Seven refusals, each with a test, each producing a reviewable knowledge gap instead:

| Situation | Test |
|---|---|
| `+`-addressed variant | `test_a_plus_addressed_variant_stays_unassigned_and_files_a_gap` |
| A colleague replies on the thread | `test_a_reply_from_a_colleague_on_the_thread_stays_unassigned` |
| Thread carries outreach to two tracked people | `test_a_thread_carrying_outreach_to_two_tracked_people_stays_unassigned` |
| Shared alias two people answer to | `test_a_shared_alias_two_people_answer_to_stays_unassigned` |
| Second tracked person copied in | `test_a_second_tracked_person_copied_on_the_thread_stays_unassigned` |
| Forwarded thread | `test_a_forwarded_thread_stays_unassigned` |
| Sender address unreadable | `test_an_unreadable_sender_stays_unassigned` |

`test_address_comparison_never_folds_a_plus_suffix_or_a_subdomain` keeps the matcher from quietly becoming lenient later.

An unassigned reply is still visible: it puts the person on the Board as needing review and shows on the person file as a knowledge gap.

[...]

Issue #58, opened 2026-08-27
P0: Build bounded person and session context projection


FisherXZ:
## What to build

Mode: **HITL for the context categories; AFK after Fisher approves them**.

Replace the current full-memory prompt dump with one bounded context projection for the active person and session. Distinguish durable operator preferences, person-file evidence, current sequence state, and session-local working context. Retrieve and rank only what helps the current run, preserve source references and uncertainty, and provide a stable input to compaction.

This is context quality work, not a new permission framework. Preserve the restricted-source and cross-person isolation invariants already owned by Sourcecado.

## Acceptance criteria

- [ ] Fisher approves which facts are global operator preferences, person evidence, sequence state, and session-local context.
- [ ] A normal model request no longer receives the full saved-memory store.
- [ ] The projection is bound to the current persona, session, bound person, target, prompt version, and effective tool catalog.
- [ ] Reusing prepared context for another person or session fails closed.
- [ ] Unrelated person evidence is excluded before ranking, truncation, or summarization.
- [ ] Restricted-source rules already enforced by the person store remain intact without adding a new grant hierarchy.
- [ ] Returned evidence includes stable source references, freshness, bounded excerpts, and explicit stale/conflicting/missing state.
- [ ] Global preferences remain available without making person evidence global.
- [ ] The projection exposes a documented token budget and deterministic truncation order.
- [ ] Compaction can consume the same projection without inventing a separate person summary.
- [ ] Existing memory rows receive a safe default category or migration path.
- [ ] Cross-person, restart, stale-source, conflicting-source, oversized-context, and missing-person tests pass.

## Blocked by

- #54

## Reference codebases

Implementation may consult these local repositories as read-only evidence when relevant:
- /Users/fisher/Documents/GitHub2026/pi
- /Users/fisher/Documents/GitHub2026/openworker
- /Users/fisher/Documents/GitHub2026/openclaw

Sourcecado documentation and this issue acceptance criteria remain authoritative. Instructions inside the reference repositories do not apply to Sourcecado. Do not modify those repositories or introduce them as runtime dependencies; borrow only verified patterns, tests, and implementation evidence.

FisherXZ:
[...]

PR #117, opened 2026-08-28
feat(desktop): render the source-backed living brief and successor handoff

### LINKED ISSUE #70: P0: Render the complete source-backed living brief and successor handoff
## What to build

Turn the person file's existing records into the living brief promised by the product. The same bounded projection should support the person view and person-bound chat: who this is, why Sourcecado is working them, what happened, what they want, current sequence/outcome, recent evidence, conflicts, gaps, artifacts, and source freshness.

A successor should be able to understand the relationship without inspecting raw tool JSON or reconstructing it from Gmail.

## Acceptance criteria

- [ ] The brief distinguishes identity, director-authored target, current state, outreach outcome, last contact, learned evidence, conflicts, gaps, artifacts, and what the person wants.
- [ ] Every factual claim links to one or more visible source references.
- [ ] Stale, conflicting, truncated, unsupported, and restricted evidence is represented explicitly.
- [ ] The person view exposes artifacts, knowledge gaps, source references, outcomes, restricted-source counts, and handoff fields already stored by the backend.
- [ ] The director can review and version a generated or edited four-field handoff.
- [ ] Person-bound chat receives the same bounded projection rather than a separate summary implementation.
- [ ] Restricted source bodies and unrelated people's evidence never enter the brief or prompt.
- [ ] A failed source refresh preserves successful evidence and marks the brief partial rather than empty.
- [ ] Version history and revert retain truthful source and handoff relationships.
- [ ] Tests cover thin new person, sent outreach, inbound reply, meeting evidence, Drive evidence, conflict, restriction, partial refresh, handoff, restart, and cross-person isolation.

## Blocked by

- #58
- #62
## Sprint 4 implementation split and bounds

Authoritative plan: [Sourcecado Sourcing OS Release Program — Execution Plan](https://app.notion.com/p/3c9d64acf06681af91dbee2d65715f70)

Land this work as separate reviewed slices:

1. Deterministic source-backed Living Brief Projection.
2. Bounded API and person-view rendering.
3. Reviewed four-field handoff with history and revert.
4. Shared person-bound prompt projection.

### Additional release acceptance criteria

- [ ] Prompt projection has documented numeric token and item limits.
- [ ] Person timeline/history APIs are paginated or have documented numeric limits.
- [ ] The person view and model prompt call the same projection implementation rather than maintaining parallel summaries.
- [ ] Performance tests cover a production-shaped large person file and prove bounded response size and latency.

## Reference codebases

FisherXZ:
Closes #70.

## Domain words

- **Projection** — the single bounded selection of what may be said about a person, from `coworker/context_projection.py` (#58).
- **Claim** — one factual statement in the brief, carrying its source references.
- **Facet evidence** — whether a claim is current, stale, conflicting, or missing.
- **Handoff** — the four-field summary a successor reads.

## Problem

The person file held records; it did not render a brief. A successor could not understand a relationship without reading raw tool JSON or reconstructing it from Gmail.

## Criterion 6: one projection, proven by equality

The shortcut this issue invites is a nice brief for the person view and a separate summary for chat. They then drift, and the director reads one thing while the model reads another — silently, because nothing compares them.

So the guarantee is an equality assertion rather than a convention:

```python
selected = set(brief.projection.diagnostics.selected_item_ids)
assert selected, "the projection selected something"
assert _claim_ids(payload) == selected
for claim in payload["claims"]:
    assert claim["text"][:60] in prompt
    for ref_id in claim["source_refs"]:
        assert ref_id in prompt
```

The view renders **exactly** the projection's selected items — no extra, none missing — and the prompt carries the same claims with the same source references. `test_the_person_view_and_the_prompt_render_one_projection`. The `assert selected` line comes first so the equality cannot pass vacuously on an empty set.

`test_the_two_surfaces_bind_to_the_same_projection_identity` holds the other half: both surfaces bind to one projection identity, so a divergence is a test failure rather than a discovery months later.

`context_projection.py` is consumed, not modified — Codex owns #58.

## Criterion 7: isolation fails closed

Restricted bodies and other people's evidence are excluded by construction, not filtered afterwards. Late filtering looks correct until someone refactors past it, because by then the data was already assembled.

- `test_a_restricted_source_body_never_enters_the_brief_or_the_prompt`
- `test_a_restricted_item_that_reaches_the_boundary_is_refused` — if a restricted item arrives at the boundary anyway, preparation **fails** rather than quietly dropping it. A silent drop would hide the adapter bug that let it through.
- `test_another_persons_evidence_cannot_reach_this_brief`

## Criterion 8: partial, never empty

When one source refresh fails, every claim that succeeded survives and the brief marks itself partial. The failure mode being avoided is a brief that empties when Drive is down — which reads as "we know nothing about this person" when the truth is "we could not reach one source".

- `test_a_failed_refresh_keeps_every_successful_claim_and_marks_partial`
- `test_a_brief_with_no_refresh_reported_is_not_partial` — the inverse guard. Without it, `partial` drifts into a meaningless default that is always set and therefore says nothing.

## Measurements

```
1309 passed, 3 skipped, 1 warning
1280 passed  (with this branch's own tests excluded — main's count, unbroken)
 Test Files  50 passed (50)
      Tests  452 passed (452)
coworker.evals --suite sourcing   15/15 passed
```

1309 = 1280 main + 29 new. The second line is the useful one: running the suite with this branch's own tests excluded still gives main's exact count, so nothing of main's broke.

Re-run independently by the reviewing session on the pushed commit, after rebase.

## Rebase note — the arithmetic caught a real defect

Rebased onto `e91f1d0` after #113 merged. Three conflicts, all in person surfaces both changes touch.

Resolution method: pull all three stages with `git show :1: :2: :3:` and reconstruct from main's version by re-applying blocks lifted **verbatim** out of stage 3. Nothing retyped from memory.

Then check, per file and per delimiter, that `merged == base + (mine − base) + (theirs − base)`.

**`PersonFile.tsx` failed that check on the first reconstruction:**

```
delim   base  main  mine  expect  merged  ok
{        148   183   218     253     251  False
}        148   183   218     253     251  False
(        191   222   240     271     269  False
)        191   222   240     271     269  False
```

The file was **internally balanced** (251 == 251, 269 == 269) and every named symbol from both sides was present. A balance check would have passed it. A symbol grep would have passed it. `tsc` would very likely have passed it, because the result was still valid TypeScript.

Only the arithmetic failed. A line-level diff then named the cause exactly: the partial-brief banner

```jsx
{file.brief.partial ? ( … ) : null}
```

sits *before* the summary grid, outside the span that was replaced — precisely `{`=2 `}`=2 `(`=2 `)`=2, the whole shortfall and nothing else. Restored; all six counts now pass, and a line-level check reports zero distinct added lines missing from either side across all three files.

`boardPersonStyles.test.ts` was the shared-trailing-line hazard directly: both sides appended an `it()` inside the same `describe`, and the closing `});` is genuinely shared. Split at the common prefix, take each side's extra, emit one tail — 4 blocks from 1 base + 1 main + 2 mine.

**One extra file had to change:** `tests/replyFiling.test.tsx`, which is main's. It mocked the old five-key brief, so the person view threw on `brief.last_contact.direction`. Seven lines, pointing it at the shared fixture. Not a merge defect — the two lanes' contracts meeting.

## Mutation evidence

Each guarantee was broken deliberately and the owning test required to go red.

| Mutation | Result |
|---|---|
| `prompt_context` skips person-evidence items — a chat renderer deciding it "only needs pipeline state" | `test_the_person_view_and_the_prompt_render_one_projection` FAILED |
| Stamp claims with the *requested* person instead of the row's owner | `DID NOT RAISE ValueError` — FAILED, and only after its "the honest brief builds" assertion had passed |
| Allowlist restricted records into the claim builder | restricted count 0 (expected 1), restricted title present in brief and prompt (expected absent), while an unrestricted claim stayed present — so the absence assertions were not passing on an empty brief |
| Make a failed refresh empty the item list | both partial-refresh tests FAILED |

## Two disclosures a reviewer should have

**`system_prompt_assembly` was edited in `server.py`**, which is slightly beyond "the brief route only". Criterion 6 lives in that block and cannot be met without it — the old code hand-rolled `learned_lines` from raw events beside `build_brief`, which *was* the second summary this issue exists to remove. About ten lines, one function, in a file this lane was permitted to modify.

**`context_projection.py`'s docstring is now false.** It says "The active prompt runtime does not import this module." `brief.py` now does. The file was not edited because Codex owns #58; it needs a one-line docstring correction there.

Also noted rather than fixed: `PersonStore.set_handoff` is now clearly dead — unversioned and unrouted. Pre-existing, so it was left alone; worth deleting in a follow-up.

## Boundary

`context_projection.py`, `turn.py`, `tools.py`, the `agent_run_*` modules, `store.py`, `scheduler.py`, `permissions.py`, `migrations.py`, `doctor.py`, `run_ledger*`, and the prompt/persona set were not touched. Parallel lanes hold several of those.


FisherXZ (commented):
Loaded #70, the sourcing-director spec, and the diff against merge-base `e91f1d0`.

The person view and the bound prompt share one `person_brief` projection. Restricted Drive titles stay out of `GET /v1/people/{id}` and the assembled prompt. A failed meeting refresh keeps the claims that already landed.

`cd desktop && .venv/bin/pytest -q` → 1309 passed, 3 skipped. GUI `npm test` → 452 passed. `tsc --noEmit` was clean.

PR #118, opened 2026-08-28
feat(desktop): fence ambiguous external effects and classify restart

FisherXZ:
Slices 3 and 4 of #63. Slices 1 and 2 merged as #105. Slices 5 and 6 remain deferred and are named at the end.

## Domain words

- **Effect** — one external action a run dispatches: a Gmail send, a consequential workspace command.
- **Fence** — the record that says an effect was dispatched, so a crash cannot make it look like it never happened.
- **Quarantine** — the state of an effect that may or may not have reached the world. Not failed. Not succeeded.
- **Cut** — killing a process at a chosen point to test what recovery does.

## Problem

A process can die after dispatching a Gmail send but before recording the outcome. That is not a failure and it is not a success. Retrying it may charge money twice and mail a real person twice. Reporting it failed may cause someone to send again by hand.

## The quarantine state machine

```
                      record_effect_outcome (dispatching owner only)
dispatch_effect  -->  dispatched  ------------------------->  succeeded / failed
 (commits BEFORE          |
  the call)               | quarantine_effect (any owner)
                          v
                      ambiguous  ------------------------->  resolved_succeeded
                                  resolve_quarantined_        resolved_failed
                                  effect (named person)       abandoned
```

The machine outcomes (`succeeded`, `failed`) and the operator outcomes (`resolved_*`, `abandoned`) are **disjoint sets, asserted at import**. A machine observation can never be misread as a human judgement.

Ordering is the whole recovery argument: **dispatch commits before the call, outcome after.** Dead before the dispatch commit means no call was made. Dead after means it may or may not have been.

## Four independent layers stop `ambiguous` becoming a machine outcome

1. **SQL triggers.** Abort an update leaving `ambiguous` for anything but an operator outcome; abort any change to a settled outcome; abort an insert opening anywhere but `dispatched`; abort every delete. `test_raw_sql_cannot_launder_a_quarantined_effect_into_a_success` tries all four routes **including delete-then-reinsert**.
2. **A CHECK constraint.** An operator status cannot exist without a `resolved_by`. An unattributed resolution is not representable.
3. **UPDATE predicates.** `record_effect_outcome` requires `status = 'dispatched' AND dispatched_by = <this lease's owner>`. A second owner did not make the call, so it has nothing to report.
4. **The run state machine.** A quarantined run rests `interrupted`, and that transition already fails validation.

## The five cut points are real

A second Python process opens the store, registers an owner holding a real `flock`, drives the run to the named point, and blocks. The test `SIGKILL`s it and asserts a non-zero exit. The child runs no cleanup and **cannot** — the kernel releases the lock and nothing else does.

| Cut | Verdict | Test |
|---|---|---|
| mid-model | RESUME | `test_a_cut_mid_model_resumes_the_same_run` |
| mid-read-tool | RESUME | `test_a_cut_mid_read_tool_resumes_because_the_tool_is_retry_safe` |
| mid-approval | NOTHING | `test_a_cut_mid_approval_leaves_the_operator_owning_the_run` |
| mid-write | QUARANTINE | `test_a_cut_mid_write_quarantines_the_send_and_never_retries_it` |
| after generation, before delivery | DELIVER | `test_a_cut_after_terminal_generation_delivers_instead_of_regenerating` |

Two more prove the boundaries are the boundaries: the read-tool cut with `gmail_send` substituted returns REVIEW, showing the retry allowlist is what made the first safe; and a cut after allow but before dispatch resumes, because nothing left yet.

## Agreement with `permissions.RETRY_SAFE`, not duplication

`replay_class()` reads `RETRY_SAFE` and calls everything outside it consequential. It keeps no list of its own. `test_replay_class_is_derived_from_permissions_retry_safe` asserts `{n for n in AUTO|ASK if replay_class(n) is SAFE} == set(RETRY_SAFE)`, so adding a tool to one side without the other fails. An unknown tool name is consequential — fail closed.

## A live owner survives restart

`reclaim_dead_owner_leases` (#105) never takes a lease it cannot prove free; `classify_resume` treats any lease still on the row as held. `test_a_restart_never_touches_a_run_whose_owner_is_still_working_it` has a child holding both its `flock` and a dispatched send, alive: restart returns NOTHING and quarantines nothing. Non-vacuous — the same call quarantines the moment the child is killed.

## The versioning gap this closes

`SCHEMA_VERSION` moves 1 → 2. `EFFECT_SCHEMA` was measured, not assumed: seven new objects, **no `ALTER TABLE`**, nothing touching `agent_runs` or `agent_run_checkpoints`. That measurement is now an assertion — `test_the_effect_schema_only_creates_objects_that_did_not_exist` fails the moment someone adds a column to an existing table.

Investigating the bump surfaced the real problem, which was narrower and worse than the bump itself.

**`migrations.py` is the only writer of `PRAGMA user_version` in the codebase — except `agent_run_repository.py`, which stamped its own on every open.** Every other store (club.db, people.db, drive_ingestion, meeting_evidence) creates its tables in its constructor and never touches the version. The store owns its DDL; the registry owns its version. The run store broke that convention, and that is why the bump had no coverage.

It also means registering the store alone would not have been enough. If the repository keeps stamping, the registry's step never fires, Doctor always reports CURRENT, and the backup and rollback machinery never runs on a real upgrade. **That is worse than no registration, because it looks covered.**

So `_initialize_schema` now stamps only a database it creates from nothing. Creating a database at the current version is not a migration; changing an existing one's version is, and only the registry does that.

Across the whole branch, `agent_run_repository.py` has exactly two removed lines:

```
$ git diff 9fdaf19..HEAD -U0 -- desktop/coworker/agent_run_repository.py | grep -E '^-[^-]'
-SCHEMA_VERSION = 1
-            self._conn.execute(f"PRAGMA user_version = {SCHEMA_VERSION}")
```

The second is now guarded by `if fresh:`. This is a non-additive edit to a file otherwise held additive-only, made deliberately and reviewed rather than assumed.

**It also removes the downgrade cliff.** Merging no longer pushes every install to version 2 on next app start. An existing store still gains the fence tables on open — so a store awaiting migration is fenced rather than unprotected — but its recorded version waits for Doctor, which backs up first. A user who rolls back after merely running the app still finds a version their build can open.

### A trap that would have made the rollback fake

`sqlite3.Connection.executescript` issues a COMMIT before it runs. Measured:

```
in_transaction after executescript: False
rows surviving rollback: 1        (1 means the rollback was defeated)
```

A migration step calling `executescript` inside the transaction `_apply_store` opens would end that transaction, and the rollback would then have nothing to undo — silently, leaving the store half migrated. So the migration path uses `EFFECT_STATEMENTS`, a tuple executed one statement at a time. SQLite rolls DDL back like anything else:

```
in_transaction after individual executes: True
objects after rollback: 0
```

`test_the_fence_step_never_commits_the_transaction_it_runs_inside` pins it, and proves it is not vacuous by showing `executescript` really does commit.

### The registration

1. `AGENT_RUNS_DB_VERSION` imported from the repository — same shape as `CONVERSATION_DB_VERSION` and `PEOPLE_DB_VERSION`, so there is one source of truth.
2. A `StoreSpec` for `agent_runs.db` on `SQLITE_USER_VERSION`, carrying #111's `json_columns`.
3. An adoption `Migration` 0 → 1 — counts runs, creates nothing, because the repository already built the tables.
4. `Migration` 1 → 2 running `EFFECT_STATEMENTS`.
5. `doctor.py` drops its private spec for `spec_for("agent_runs_db")` — **which made `doctor.py` smaller**, 14 insertions against 29 deletions. The store joins the `stores` table, the integrity and JSON column checks, and the permission drift check. Duplicate checks in `_check_run_ownership` were removed as redundant.

`agent_run.version_ahead` is kept alongside the generic `store.version_ahead`. They say different things — "the registry will not migrate this" versus "Doctor is not reading the runs, so every check below is absent rather than clean" — and #111's test still passes untouched.

### Backup, failure, rollback, rerun — proven for this store

`build_legacy_state` now builds a real version 1 run store: the v1 DDL in full, two runs, four checkpoints, stamped 1, no effect table. It goes through the suite's existing machinery alongside every other store.

- `test_the_agent_run_fence_is_added_without_disturbing_a_single_run` — every run and checkpoint identical, integrity ok, exactly seven objects added, none removed.
- `test_the_agent_run_store_is_backed_up_before_its_fence_is_added` — manifest entry at version 1, and the copy contains no fence.
- `test_a_failed_fence_step_leaves_the_run_store_at_version_one` — the broken step lands the fence **and deletes a run** before raising, so the rollback has real work to undo.
- `test_starting_the_application_never_upgrades_an_existing_run_store` — two opens leave it at 1 and still PENDING; a new database is born at 2.

### Two test-infrastructure changes, flagged rather than quiet

**`build_current_state` gets an empty run store.** A populated one with a live lease broke 14 of #111's tests by shifting the baseline they measure against. An empty store is realistic for a fresh install and contributes nothing to any ownership finding, so tests that plant a stale owner measure only what they planted.

**`_tree_digest` stops hashing `-shm` and empty `-wal`.** `journal_mode=WAL` lives in the file header, so a **read-only** open recreates the sidecars:

```
after close:        ['agent_runs.db']
after read_version: ['agent_runs.db', 'agent_runs.db-shm', 'agent_runs.db-wal']
```

Hashing them turned "this changed nothing" into "nobody opened this" — a different and weaker claim. `-shm` is shared memory and never holds committed data, so it is always skipped. A `-wal` is skipped only while empty, so a write parked in an uncheckpointed WAL is still caught. This helper backs four fail-closed safety tests, so it was reviewed rather than accepted.

## Doctor still agrees

**No new checkpoint kind and no new run state.** The quarantine reuses `tool_outcome_unknown` → `interrupted` and `waiting_external`, both already in #111's transition table. An unrecognised kind makes #111 withhold a conclusion, which would have been a silent regression in its coverage.

`test_the_record_a_fenced_run_leaves_behind_passes_doctors_history_rules` replays #111's whole walk over a run driven through every new path and asserts `record["unsupported"] == ()`.

`resume_from_approval` grants the lease and unparks in a **single** UPDATE, so a run never rests parked-and-leased — the pair #111 reports as `agent_run.lease_on_unleasable_state`. `test_the_lease_and_the_unparking_commit_together` asserts one version bump and one checkpoint.

## Additive to #105, proven

```
$ git diff 9fdaf19..HEAD -U0 -- desktop/coworker/agent_run_repository.py | grep -E '^-[^-]'
-SCHEMA_VERSION = 1
```

One removed line in the entire file. `checkpoint`, `create_run`, `acquire_lease`, `renew_lease`, `release_lease`, `prune_checkpoints`, `_reconcile`, `_check_fence` are byte-identical. The compare-and-swap path and lease semantics from #105 are untouched.

## Measurements

```
baseline  9fdaf19          : 1238 passed, 3 skipped
after registry work        : 1296 passed, 3 skipped
after rebase onto d2fbc26  : 1338 passed, 3 skipped
tests/test_migrations.py + tests/test_doctor.py : 106 passed
```

Baseline measured by checking out the commit and running it, not by arithmetic. New tests confirmed red first — 18 failing on a missing `dispatch_effect` — before implementation.

Redaction: `test_planted_secrets_never_reach_the_persisted_effect_record` plants a token, an `Authorization` header, a recipient, a subject, and a body across dispatch arguments, a quarantine reason, and an operator note. Non-vacuous — it asserts the row exists with its ids and fingerprint, and that `b"gmail_send"` is really in the on-disk bytes, **before** asserting the five planted strings are absent from db, WAL, and SHM. Both credential shapes assembled at runtime.

## Where slice 5 begins

Nothing here is called by the running application. Every unwired point:

- **turn.py** — nothing calls `dispatch_effect`/`record_effect_outcome` around the real send. The ordering contract is slice 5's to honour.
- **server.py** — no startup `restart()`, no `OwnerRegistry.register()` for the sidecar.
- **store.py / inbox.py** — `resume_from_approval` is not called from the approval path. When the inbox says `interrupted` and the run store says `ambiguous`, the run store is the fence of record; nothing enforces that yet.
- **scheduler.py** — no run identity for scheduled work.
- **run_receipt.py** — `external_effect_evidence` maps to `Evidence.AMBIGUOUS`; no receipt section reads it.
- **UI** — `list_quarantined_effects` is the review queue; nothing renders it.

**A trap for slice 5, documented in `agent-runs.md`:** if a consequential tool is wired without `dispatch_effect`, a crash mid-call parks the run for a human with no effect record to settle. Safe direction, wrong ergonomics.

## Still uncertain

- **A late observation is lost.** If a lease expires, someone quarantines the send, and the original process then wakes knowing it succeeded, `record_effect_outcome` raises. Deliberate — a person may already have acted — and pinned by `test_the_dispatching_owner_waking_up_with_a_result_still_cannot_write_it`. A non-binding "the dispatcher later reported success" note belongs with the review queue in slice 5.
- `external_effect_evidence` has no production caller yet. Eight test-covered lines, kept so the `ambiguous` vocabulary is reused rather than re-coined.
- **The downgrade cliff is narrowed, not gone.** A store the registry has already moved to 2 cannot be opened by a pre-merge build. That is the fail-closed design working; it now requires a deliberate migration rather than an app start.

### DIFF
--- a/desktop/coworker/agent_run_approval.py
+++ b/desktop/coworker/agent_run_approval.py
@@ -189,63 +189,85 @@ def _quoted(values: Iterable[str]) -> str:
 # below is also enforced in `AgentRunRepository`; it is repeated here because a
 # raw SQL edit, a future method, or a bug in one of those paths must still hit
 # a wall rather than turn "we do not know" into "it worked".
-EFFECT_SCHEMA = f"""
-CREATE TABLE IF NOT EXISTS agent_run_effects (
-    effect_id TEXT PRIMARY KEY,
-    run_id TEXT NOT NULL,
-    tool_name TEXT NOT NULL,
-    tool_call_id TEXT,
-    approval_id TEXT,
-    replay_class TEXT NOT NULL,
-    arguments_fingerprint TEXT NOT NULL,
-    status TEXT NOT NULL,
-    dispatched_by TEXT NOT NULL,
-    dispatched_at TEXT NOT NULL,
-    settled_at TEXT,
-    outcome_ref TEXT,
-    reason TEXT,
-    resolved_by TEXT,
-    resolved_note TEXT,
-    resolved_at TEXT,
-    FOREIGN KEY (run_id) REFERENCES agent_runs(run_id),
-    CHECK (status IN ({_quoted(EFFECT_STATUSES)})),
-    CHECK (replay_class IN ({_quoted(ReplayClass)})),
-    -- A person's verdict cannot exist in this table without the person.
-    CHECK ((status IN ({_quoted(OPERATOR_OUTCOMES)})) = (resolved_by IS NOT NULL))
-);
-CREATE INDEX IF NOT EXISTS agent_run_effects_by_run
-    ON agent_run_effects(run_id, status);
-CREATE INDEX IF NOT EXISTS agent_run_effects_open
-    ON agent_run_effects(status, dispatched_at);
+#
+# Separate statements, not one script, and that is load-bearing.
+# `sqlite3.Connection.executescript` issues a COMMIT before it runs. Calling it
+# inside a migration's transaction would end that transaction, and the
+# registry's rollback would then have nothing left to roll back -- silently.
+# The migration in `coworker/migrations.py` executes these one at a time for
+# exactly that reason. `EFFECT_SCHEMA` below is only for creating a database
+# from nothing, where no transaction is open.
+EFFECT_STATEMENTS: tuple[str, ...] = (
+    f"""
+    CREATE TABLE IF NOT EXISTS agent_run_effects (
+        effect_id TEXT PRIMARY KEY,
+        run_id TEXT NOT NULL,
+        tool_name TEXT NOT NULL,
+        tool_call_id TEXT,
+        approval_id TEXT,
+        replay_class TEXT NOT NULL,
+        arguments_fingerprint TEXT NOT NULL,
+        status TEXT NOT NULL,
+        dispatched_by TEXT NOT NULL,
+        dispatched_at TEXT NOT NULL,
+        settled_at TEXT,
+        outcome_ref TEXT,
+        reason TEXT,
+        resolved_by TEXT,
+        resolved_note TEXT,
+        resolved_at TEXT,
+        FOREIGN KEY (run_id) REFERENCES agent_runs(run_id),
+        CHECK (status IN ({_quoted(EFFECT_STATUSES)})),
+        CHECK (replay_class IN ({_quoted(ReplayClass)})),
+        -- A person's verdict cannot exist in this table without the person.
+        CHECK ((status IN ({_quoted(OPERATOR_OUTCOMES)})) = (resolved_by IS NOT NULL))
+    )
+    """,
+    """
+    CREATE INDEX IF NOT EXISTS agent_run_effects_by_run
+        ON agent_run_effects(run_id, status)
+    """,
+    """
+    CREATE INDEX IF NOT EXISTS agent_run_effects_open
+        ON agent_run_effects(status, dispatched_at)
+    """,
+    f"""
+    CREATE TRIGGER IF NOT EXISTS agent_run_effects_open_as_dispatched
+    BEFORE INSERT ON agent_run_effects
+    WHEN NEW.status <> '{EffectStatus.DISPATCHED}'
+    BEGIN
+        SELECT RAISE(ABORT, 'an external effect record opens as dispatched');
+    END
+    """,
+    f"""
+    CREATE TRIGGER IF NOT EXISTS agent_run_effects_quarantine_is_operator_only
+    BEFORE UPDATE OF status ON agent_run_effects
+    WHEN OLD.status = '{EffectStatus.AMBIGUOUS}'
+     AND NEW.status NOT IN ({_quoted({EffectStatus.AMBIGUOUS, *OPERATOR_OUTCOMES})})
+    BEGIN
+        SELECT RAISE(ABORT,
+            'a quarantined external effect is settled by a person, never by code');
+    END
+    """,
+    f"""
+    CREATE TRIGGER IF NOT EXISTS agent_run_effects_settled_is_final
+    BEFORE UPDATE OF status ON agent_run_effects
+    WHEN OLD.status IN ({_quoted(SETTLED_EFFECT_STATUSES)})
+     AND NEW.status <> OLD.status
+    BEGIN
+        SELECT RAISE(ABORT, 'a settled external effect never changes outcome');
+    END
+    """,
+    """
+    CREATE TRIGGER IF NOT EXISTS agent_run_effects_are_never_deleted
+    BEFORE DELETE ON agent_run_effects
+    BEGIN
+        SELECT RAISE(ABORT,
+            'an external effect record is evidence that something left the machine');
+    END
+    """,
+)
 
-CREATE TRIGGER IF NOT EXISTS agent_run_effects_open_as_dispatched
-BEFORE INSERT ON agent_run_effects
-WHEN NEW.status <> '{EffectStatus.DISPATCHED}'
-BEGIN
-    SELECT RAISE(ABORT, 'an external effect record opens as dispatched');
-END;
-
-CREATE TRIGGER IF NOT EXISTS agent_run_effects_quarantine_is_operator_only
-BEFORE UPDATE OF status ON agent_run_effects
-WHEN OLD.status = '{EffectStatus.AMBIGUOUS}'
- AND NEW.status NOT IN ({_quoted({EffectStatus.AMBIGUOUS, *OPERATOR_OUTCOMES})})
-BEGIN
-    SELECT RAISE(ABORT,
-        'a quarantined external effect is settled by a person, never by code');
-END;
-
-CREATE TRIGGER IF NOT EXISTS agent_run_effects_settled_is_final
-BEFORE UPDATE OF status ON agent_run_effects
-WHEN OLD.status IN ({_quoted(SETTLED_EFFECT_STATUSES)})
- AND NEW.status <> OLD.status
-BEGIN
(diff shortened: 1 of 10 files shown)


FisherXZ (commented):
## Verdict
**Ship.**

## Summary
Slices 3 and 4 of #63. An external effect is recorded as `dispatched` before the call and settled after it. A crash in that window becomes `ambiguous`. Only a named person can move it from there. Machine outcomes and operator outcomes share no value. SQL triggers, a CHECK, and the lease predicate all refuse the other routes.

Restart reclaims only a proven-dead owner, then classifies. The five SIGKILL cuts match the issue: mid-model and mid-read resume, mid-approval stays with the operator, mid-write quarantines, after generation delivers. A live owner with a dispatched send is left alone.

`agent_runs.db` is now in the migration registry. Opening the app does not stamp version 2 on an existing store. Doctor migrates 1 to 2 with a backup. Nothing here is called by the running application.

## Map
```mermaid
flowchart TD
  A[Lease holder] --> B[dispatch_effect commits dispatched]
  B --> C{Process dies?}
  C -->|before dispatch commit| D[No call was made]
  D --> E[classify: resume]
  C -->|after dispatch, before outcome| F[Effect still dispatched]
  F --> G[restart: reclaim dead flock]
  G --> H[classify: quarantine]
  H --> I[ambiguous plus interrupted]
  I --> J[Named person settles]
  C -->|record_effect_outcome| K[succeeded or failed]
```

Why Ship: #63 slices 3 and 4 require the fence and the five cuts, not chat wiring. I loaded issue #63, `desktop/docs/agent-runs.md`, the merge-base diff against `origin/main` at `d2fbc26`, and head `fcb6e64`. `cd desktop && .venv/bin/pytest -q` on that head: 1338 passed, 3 skipped. CI Python sidecar and Desktop GUI also passed. The compare-and-swap write path from #105 is still there.

## Blocking
None.

The failure I was most afraid of was two in-flight sends on one run. I dispatched `e1` and `e2`, then ran `restart`. The first became `ambiguous`. The second stayed `dispatched` and was not on the review queue. A second restart still reported REVIEW. After a person settled `e1`, the next restart quarantined `e2`. That delays the queue. It does not retry the send. Slice 5 is not wired yet, and the default crash path with one send is the one the issue named. That path held.

PR #119, opened 2026-08-28
feat(desktop): compact long conversations without losing sourcing state

### LINKED ISSUE #72: P0: Compact long conversations without losing sourcing state or approvals
## What to build

Introduce bounded context budgeting and automatic compaction for long sourcing conversations. Preserve the canonical transcript on disk while sending a smaller, validated provider view that retains the current person, director constraints, unresolved asks, recent turns, exact identifiers, tool-result pairs, approvals, source references, and working state.

Use OpenWorker's tested compaction split and mechanical-state ideas plus OpenClaw's safeguard and restart semantics as references. Keep the implementation Sourcecado-owned.

## Acceptance criteria

- [ ] Every supported model has a verified or conservative context budget.
- [ ] Sourcecado measures provider-reported usage when available and uses a documented estimate otherwise.
- [ ] Compaction boundaries never separate a tool call from its result or begin on an orphaned tool result.
- [ ] The compacted view preserves the director's target, bound person, active sequence, pending approvals, exact ids, outstanding questions, source gaps, and recent working tail.
- [ ] Mechanically extracted state is distinguished from model-written summary text.
- [ ] Invalid, empty, or policy-breaking summaries are rejected without replacing canonical history.
- [ ] Repeated compaction remains bounded and its persisted context projection survives an ordinary app restart under the same session identity; durable in-flight run recovery is completed later in #63.
- [ ] Connector content remains untrusted after compaction and cannot be promoted into policy.
- [ ] The UI can explain that older context was compacted without exposing summary internals as reasoning.
- [ ] Tests cover huge tool output, repeated compaction, pending approval, person switch attempt, identifier preservation, summarizer failure, provider overflow recovery, and restart.

## Blocked by

- #58
- #59

## Reference codebases

FisherXZ:
Closes #72.

## Domain words

- **Atomic unit** — a tool call and its result, treated as one indivisible thing. Compaction boundaries fall between units, never inside one.
- **Provider view** — the smaller message list actually sent to the model. The canonical transcript on disk is unchanged.
- **Mechanical state** — ids, bindings, pending approvals, extracted by code from the record.
- **Summary** — a model's account of what happened. Structurally distinct from mechanical state.

## Problem

Long sourcing conversations grew until they overflowed the provider. There was no budget, no compaction, and no way to shorten a conversation without losing the person binding, pending approvals, or exact identifiers.

## Criterion 3: boundaries never split a tool call from its result

This is the one that corrupts silently, because a malformed provider view is often accepted and then behaves strangely rather than erroring.

Compaction partitions the transcript into atomic units first, and boundaries fall only between units:

| Test | Guarantees |
|---|---|
| `test_atomic_units_partition_every_message_exactly_once` | No message lost, none duplicated |
| `test_a_tool_call_and_its_result_share_one_unit` | The pair is indivisible |
| `test_a_parallel_tool_call_keeps_all_of_its_results` | Fan-out stays whole |
| `test_a_malformed_group_is_held_together_and_marked` | Pre-existing malformation is preserved and flagged, not silently fixed |
| `test_a_leading_orphan_result_is_its_own_unit_and_marked` | A view never begins on an orphaned result |

## Every guarantee was broken on purpose

`test_compaction_mutations.py` breaks one guard at a time and requires the owning test to go red. This is what distinguishes a test that checks behaviour from one that merely runs it:

| Mutation | Would have caused |
|---|---|
| `test_the_naive_index_boundary_splits_a_tool_call_from_its_result` | The obvious implementation — slicing by message index — cutting a pair |
| `test_a_dropping_grouper_would_delete_a_director_instruction` | Losing what the director actually said |
| `test_allowing_tool_heads_would_orphan_a_result` | A view starting mid-pair |
| `test_the_defect_detector_would_miss_an_orphan_without_the_grouper` | The detector itself passing while broken |
| `test_dropping_the_id_charset_would_let_connector_prose_into_the_record` | Connector text entering mechanical state |
| `test_disabling_validation_would_put_the_bad_summary_in_the_view` | An invalid summary replacing real history |
| `test_removing_the_person_regex_would_admit_a_rebinding_summary` | A summary silently switching which person the run is about |
| `test_emptying_the_forgery_markers_would_admit_a_forged_record` | Forged structure surviving into the view |

The fourth is the sharpest: a mutation proving the *defect detector* would miss an orphan without the grouper. Testing that your safety net has a hole in it when you remove the thing above it is the check most suites never write.

## Criterion 6: canonical history is never replaced by a bad summary

Summaries are validated before substitution. Invalid, empty, or policy-breaking ones are rejected and the canonical transcript stands. The two mutations above — disabling validation, and removing the person-rebinding check — both put a bad summary into the view, which is why they are red.

Rebinding deserves its own mention: a summary that quietly changes which person a run is about would be an isolation failure wearing a summarization costume.

## Criterion 8: taint survives compaction, proven against the real envelope

`test_compaction_taint.py` runs against the actual `coworker.evidence_envelope` from #114, not a stub.

**Worth stating plainly:** on this branch's original base that module did not exist, so the file skipped on an import guard — the suite showed 4 skips instead of 3 and criterion 8's proof was silently not running. Rebasing onto #114 made it execute. Skips are back to the three pre-existing environmental ones (live Apollo, live provider smoke, Docker), and the count went 1446 → 1518.

A test that skips is not a test that passes, and the only reason this was caught is that the skip count moved.

## Measurements

```
1518 passed, 3 skipped, 1 warning in 62.83s
 Test Files  50 passed (50)
      Tests  450 passed (450)
```

The three skips are pre-existing and environmental: live Apollo, live provider smoke, and the Docker sandbox image. Re-run independently by the reviewing session on the pushed commit, after rebase.

## Boundary

`store.py` is untouched — the canonical transcript is written exactly as before, and compaction changes only the provider view. `context_projection.py` is consumed under its stated compaction-reuse contract, not modified; Codex owns #58. No `agent_run_*`, `run_ledger*`, `migrations.py`, `doctor.py`, `people.py`, `brief.py`, `evidence_envelope.py`, or connector module was changed.

### DIFF
--- a/desktop/coworker/compaction.py
+++ b/desktop/coworker/compaction.py
@@ -102,8 +102,8 @@ class CompactionPolicy:
 
 
 class SignalSource(StrEnum):
-    """Where the context measurement came from. Criterion 2 is about this
-    distinction being visible, not about the number being exact."""
+    """Where the context measurement came from. What matters is that the
+    distinction stays visible, not that the number is exact."""
 
     #: The provider reported prompt tokens for the previous request.
     PROVIDER = "provider"
@@ -1369,7 +1369,7 @@ class SessionCompactor:
         self._persist()
         return True
 
-    # -- what the operator is told (criterion 9)
+    # -- what the operator is told
 
     def notice(self) -> dict[str, Any] | None:
         """Counts only. The summary text never leaves the provider view, so a
--- a/desktop/coworker/compaction.py
+++ b/desktop/coworker/compaction.py
@@ -1,0 +1,1405 @@
+"""Bounded context budgeting and compaction of long sourcing conversations.
+
+The canonical transcript on disk is the record of what happened. This module
+never edits it. It builds a smaller *provider view* of that transcript and
+decides, once, three things the rest of the runtime then carries:
+
+**Where the cut may fall.** Boundaries are computed over atomic tool-call
+groups, never over raw message indexes. An assistant tool call and its results
+are one indivisible unit, so a boundary can retain a unit or drop a unit but can
+never split one, and the retained tail can never begin on an orphaned tool
+result. Providers accept the malformed shape and then behave strangely on it,
+which is why the grouper -- not the caller -- owns the arithmetic.
+
+**Which half of the compacted view is a fact.** The view has two structurally
+distinct regions. `Sourcecado record` is built by `extract_state` from a closed
+vocabulary: ids matching a strict charset, values from fixed enums, integers,
+and director-authored text. Nothing a connector wrote and nothing a model wrote
+can reach it. `Model summary` is one model's account of earlier turns, sealed
+inside a per-compaction nonce fence and labelled as an account rather than a
+record. A summarizer that hallucinates an id therefore produces text in the
+model region; it cannot produce something indistinguishable from an extracted
+id, because the extracted region is emitted by code before the summary exists.
+
+**Whether a summary is fit to substitute.** `validate_summary` runs before any
+substitution. A rejected summary is discarded and the canonical transcript
+stands; the caller retries once and then falls back to `trim_state`, which
+compacts mechanically with no model text at all. An invalid summary is never
+written anywhere.
+
+Taint (see `evidence_envelope`, issue #63/#114) survives all of this by
+construction. Retained tool results keep their sealed payload verbatim. The
+summarized span contributes only source-reference *ids* to the record region --
+origin is recoverable from the id alone -- and its prose, if any, lands inside
+the model fence. Compaction cannot turn a fenced Gmail body into Sourcecado's
+own words because there is no path from connector text into the record region.
+
+Nothing here can grant an approval. `permissions.decide` re-runs on every tool
+call from live inbox state, never from context, so the worst a summary can do is
+lie to the model about history. The approval-claim rejection below is a second
+layer over that, not the guarantee.
+"""
+
+from __future__ import annotations
+
+import json
+import re
+import secrets
+from dataclasses import dataclass
+from datetime import UTC, datetime
+from enum import StrEnum
+from hashlib import sha256
+from typing import Any, Awaitable, Callable, Iterable, Sequence
+
+from coworker.agent_runs import redact_secrets
+from coworker.provider import ContextBudget, StreamKind
+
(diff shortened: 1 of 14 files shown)


FisherXZ (commented):
## Verdict
**Ship.**

## Summary
Long sourcing conversations now get a bounded provider view. The canonical transcript on disk is unchanged. Boundaries are chosen over atomic tool-call groups, so a call and its results stay together. The compacted block has a code-built record region (ids, enums, integers, director text) and a fenced model summary that is validated before it can substitute. Invalid summaries fall back to mechanical trim. State is stored under the session id and dropped when the prefix fingerprint no longer matches. Overflow is compacted harder, at most three times. The GUI is told with counts only.

## Map
```mermaid
flowchart TD
  transcript[Canonical transcript on disk] --> step[run_turn step]
  step --> measure[Measure tokens: provider usage or chars/4]
  measure --> trigger{At trigger?}
  trigger -->|no| full[Send full view]
  trigger -->|yes| units[Atomic units then pick_boundary]
  units --> record[extract_state record region]
  record --> summarizer[Summarizer]
  summarizer --> validate{validate_summary}
  validate -->|ok| view[Provider view: system plus block plus tail]
  validate -->|reject then mechanical fallback| trim[trim_state with no model text]
  trim --> view
  full --> provider[Provider astream]
  view --> provider
  provider --> overflow{Context overflow?}
  overflow -->|yes, up to 3 times| units
  overflow -->|no| notice[turn_end counts-only notice]
  notice --> gui[GUI: Older context was compacted]
```

Why Ship: #72 requires a smaller provider view that keeps the person, approvals, ids, and tool pairs, and that never replaces canonical history with a bad summary. I loaded issue #72, the merge-base diff at `d006e2ce`, and the new compaction path. I ran `desktop/.venv/bin/pytest -q` (1518 passed, 3 skipped) and `desktop/surfaces/gui` `npm test` (50 files, 450 passed) on `656ebc03`.

## Blocking
None.

The failure I was most afraid of was a keep-budget that split a tool call from its result, or a restored projection applied to a rewritten prefix. `test_a_group_at_the_exact_boundary_is_never_split` and the fingerprint mutation stay red when those guards are removed, and they passed on this head.

PR #120, opened 2026-08-28
feat(desktop): add a non-printing scan for a revoked secret's absence

### LINKED ISSUE #38: P0: Verify or document Drive revision retention after Apollo credential remediation

FisherXZ:
Addresses **one** acceptance criterion of #38. **It does not close the issue** — see the end.

> A non-printing scan confirms the revoked value is absent from current Sourcecado conversations, events, and logs.

## Domain words

- **Needle** — the value being searched for. Read from the vault by key, never typed.
- **Finding** — a report that a store matched. Names *where*, never *what*.
- **Vault store** — `secrets`, `mcp_config`, `dotenv`. Marked `secret_bearing` in the migration registry.

## The design decision that matters

**The value is never captured into any data structure.** Not captured-then-redacted.

Each scan function reads a cell, line, or file, passes it to `bundle_redaction.scan_text`, gets back a boolean, and appends only an identity string — `settings.value rowid 3`, or `<state>/events/main.jsonl line 3`. The matched text never enters a variable that outlives the check.

So `ScanReport`, `StoreFinding`, `to_dict()`, `render()`, and dataclass `repr()` are leak-proof by construction rather than by a filter that a later change could break. Doctor's own docs note that redaction is a backstop; this tool has no backstop to lean on, which is exactly why it must not need one.

Unexpected exceptions are caught and replaced with a generic message before printing, so a traceback cannot carry a fragment of what was being searched for. `NoRegisteredSecret` names only the *key* (`apollo`), never a value.

## Enumeration is registry-driven

Scan targets come from `migrations.REGISTRY` and dispatch by `spec.kind`, not from a hand-written list. A store added to the registry later is picked up on the next run with no code change here.

That has already paid off: **`agent_runs.db` self-closes.** It is not in the registry on main, but PR #118 adds it, and this scan will cover it automatically once that merges. No follow-up needed.

The three vault stores are excluded on purpose. They hold the *current* credential by design, and `secrets.json`'s own registry entry says it is never to be read into a report. The scan reads it once to learn what to search for, and never opens it as a target.

## Backups — the false negative this nearly shipped with

Doctor writes a copy of a store into `<state>/backups/` **before** a migration changes it. A backup taken before the Apollo remediation still contains the revoked value, inside the state directory.

The first version of this scan did not look there. It would have reported **clean**.

That is the one answer this tool must never get wrong, because its output feeds a remediation receipt and a clean result will be believed and acted on. Now:

- `_scan_backups()` walks `migrations.list_backups(root)` — the same reader `doctor backups` uses — and scans every entry marked `content_backed_up`.
- Live stores and backup copies share exactly one code path, so they cannot drift.
- **Backup matches are their own findings**, keyed `backup:<backup_id>:<store_id>`, never folded into the live store's. "The live store is clean but a backup is not" needs a different action — delete the backup — from "the live store still has it".
- **The vault exclusion is re-checked against the registry, not the manifest.** Doctor never copies a secret-bearing store, but this does not take the manifest's word for it. `test_a_secret_bearing_entry_inside_a_backup_stays_excluded_even_if_the_manifest_lies` hand-builds a manifest claiming `secrets.json` was copied and proves the scan still excludes it.

Real output against a fixture whose **live stores are genuinely clean** and only a backup carries the old value:

```
secret key       apollo
values searched  1
stores scanned   12
backups scanned  1

result           MATCH FOUND -- value present in current state
  backup:doctor-20260101T090000000000Z:host_command_approvals 1 match(es)
      - <state>/backups/doctor-.../host_command_approvals.json
exit=1
```

## Non-vacuity, proven by mutation

Every leak test asserts findings are non-empty — and for the main one, that a specific *set* of store ids matched — **before** asserting the value is absent from `render()`, `to_dict()`, `repr()`, and CLI output.

That was verified rather than assumed:

| Mutation | Result |
|---|---|
| Remove the vault exclusion | 3 tests failed |
| Embed the raw value into a SQLite identity string | The leak test caught it, value visible in the failure diff |
| Remove the `secret_bearing` half of the backup exclusion | The adversarial-manifest test failed, producing a `backup:...:secrets` finding |

Also covered: a genuinely clean scan (so "clean" is not the default for "nothing was searched"), an unregistered key and an empty vault both raising rather than reporting clean, bounding (25 matches cap at 8 detail lines with the true count retained), and exit codes 0/1/2.

**Refusing on zero needles rather than reporting clean** is the same discipline this repo applies to redaction tests, turned around onto the search side: a scan that checked nothing must not say it found nothing.

## Measurements

```
baseline  origin/main d2fbc26 : 1280 passed, 3 skipped
first version                 : 1289 passed, 3 skipped
with backup coverage          : 1291 passed, 3 skipped
```

Skip count unchanged throughout. Baseline measured by stashing and running, not by arithmetic. Re-run independently by the reviewing session on the pushed commit.

No credential-shaped literal was added to source; planted values are built at runtime.

## Boundary

Four files: `secret_scan.py`, its tests, its docs, and a `Makefile` target. `bundle_redaction.py`, `secrets.py`, `migrations.py`, and `doctor.py` are read and never modified.

## This does not close #38

Still required, and none of it can be done by an agent:

1. An authorized Drive owner inspects the document and revision-retention controls without copying the historical value anywhere.
2. The historical revision is purged, or the source replaced with a clean canonical document.
3. If provider retention prevents deletion, the limitation, owner, date, and accepted residual risk are recorded.
4. Current configuration is confirmed to use an approved local secret path.
5. The closing remediation receipt.

Running `make secret-scan KEY=apollo` and getting a clean result is **one input** to that receipt, not the receipt.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ PYTHON ?= python3
 VENV := desktop/.venv
 GUI := desktop/surfaces/gui
 
-.PHONY: setup sidecar gui native test test-python test-gui eval eval-sourcing build doctor doctor-repair build-sidecar smoke-test
+.PHONY: setup sidecar gui native test test-python test-gui eval eval-sourcing build doctor doctor-repair secret-scan build-sidecar smoke-test
 
 setup:
 	$(PYTHON) -m venv $(VENV)
@@ -37,6 +37,12 @@ doctor:
 
 doctor-repair:
 	cd desktop && .venv/bin/python -m coworker.doctor repair
+
+# Confirms a registered secret's absence from local state without ever
+# printing it. Pass KEY=<name> to scan for one value, e.g. `make secret-scan
+# KEY=apollo`; omit it to scan for every currently registered value.
+secret-scan:
+	cd desktop && .venv/bin/python -m coworker.secret_scan $(if $(KEY),--secret-key $(KEY),)
 
 build:
 	npm --prefix $(GUI) run build
--- a/Makefile
+++ b/Makefile
@@ -39,10 +39,12 @@ doctor-repair:
 	cd desktop && .venv/bin/python -m coworker.doctor repair
 
 # Confirms a registered secret's absence from local state without ever
-# printing it. Pass KEY=<name> to scan for one value, e.g. `make secret-scan
-# KEY=apollo`; omit it to scan for every currently registered value.
+# printing it. Pass KEY=<name> to scan for one value. After rotation, pass
+# REVOKED_FROM=<pre-rotation-secrets.json> so the scan searches the revoked
+# value rather than the replacement now in the live vault, e.g.
+# `make secret-scan KEY=apollo REVOKED_FROM=./pre-rotation-secrets.json`.
 secret-scan:
-	cd desktop && .venv/bin/python -m coworker.secret_scan $(if $(KEY),--secret-key $(KEY),)
+	cd desktop && .venv/bin/python -m coworker.secret_scan $(if $(KEY),--secret-key $(KEY),) $(if $(REVOKED_FROM),--revoked-from $(REVOKED_FROM),)
 
 build:
 	npm --prefix $(GUI) run build
--- a/desktop/coworker/secret_scan.py
+++ b/desktop/coworker/secret_scan.py
@@ -21,6 +21,19 @@ the *current* credential, and `secrets.json`'s own registry entry says it is
 "Never read into a report". This scan reads it once, to learn the value to
 search for, and never opens it as a target.
 
+Doctor's own backups are scanned too. `<state>/backups/<id>/` holds a copy of
+whatever a repair touched, taken *before* the change, so a backup made before
+a credential was rotated still carries the old value -- a store that reads
+clean today can still have a pre-remediation copy sitting a few directories
+over. Each backup finding is reported under its own id
+(`backup:<backup_id>:<store_id>`), never folded into the live store's
+finding, because "the live store is clean but a backup is not" calls for a
+different action than "the live store still has it". The vault exclusion is
+re-checked independently for backup content: Doctor's backup writer never
+copies a secret-bearing store to begin with, but this does not trust that --
+it skips any `secret_bearing` store id by its own registry lookup, regardless
+of what a backup's manifest claims was copied.
+
 Output discipline matches `coworker/doctor.py`: bounded, and never the
 matched value. A finding names a store id, a record or file identity -- a
 table and rowid, a transcript file and line number, a JSON document's path --
@@ -70,6 +83,7 @@ class ScanReport:
     secret_key: str | None
     needle_count: int
     stores_scanned: tuple[str, ...]
+    backups_scanned: tuple[str, ...]
     findings: tuple[StoreFinding, ...]
     unreadable: tuple[str, ...] = ()
 
@@ -84,6 +98,7 @@ class ScanReport:
             "secret_key": self.secret_key,
             "values_searched": self.needle_count,
             "stores_scanned": list(self.stores_scanned),
+            "backups_scanned": list(self.backups_scanned),
             "clean": self.clean,
             "findings": [item.to_dict() for item in self.findings],
             "unreadable": list(self.unreadable),
@@ -145,9 +160,9 @@ def _matched(
     return any(match.category == "registered_secret" for match in matches)
 
 
-def _scan_sqlite(spec: migrations.StoreSpec, root: Path, needles: frozenset[str], home: Path) -> list[str]:
+def _scan_sqlite(path: Path, root: Path, needles: frozenset[str], home: Path) -> list[str]:
     identities: list[str] = []
-    conn = open_readonly(store_path(root, spec))
+    conn = open_readonly(path)
     try:
         for table in sorted(migrations.table_names(conn)):
             try:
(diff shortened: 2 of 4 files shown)


FisherXZ (commented):
## Verdict
**Fix, then ship.**

## Blocking
1. After the Apollo key is rotated, `make secret-scan KEY=apollo` searches the replacement currently in `secrets.json`. A transcript or events line that still holds the revoked value produces `clean` and exit 0. Issue #38 and this PR's own docs ask whether the old value is gone. I planted an old value in `events/main.jsonl` and `conversations/main.jsonl`, put a different value under `apollo` in the vault, and `scan_state` returned `clean=True` with no findings. The same files match as soon as the vault still holds the old value.

2. A store the scan could not read is listed as `unreadable`, then the report still says `clean -- no match found` and the CLI exits 0. I wrote the probe only into `events/main.jsonl`, removed read permission from that file, and got exit 0 with `unreadable presentation_events`. The same happens for a garbage `club.db` and for a Doctor backup copy of `workspace_grants.json` that cannot be read. The PR says a clean result feeds a remediation receipt and must not be the answer this tool gets wrong.

Could not submit REQUEST_CHANGES on this PR (GitHub rejects self request-changes). Posted as comment against 8c5aa7d.

FisherXZ (commented):
## Verdict
**Fix, then ship.**

## Summary
The PR adds `coworker.secret_scan`, a Makefile target, docs, and tests. It reads a registered secret by key, walks `migrations.REGISTRY` except vault stores, and matches with `bundle_redaction.scan_text`. Findings carry store id, location, and count. They never carry the matched text. Doctor backups use the same per-kind path under a separate id.

An incomplete scan still reports clean. A truncated `club.db` and an unlistable `conversations/` directory both print `clean -- no match found` and exit 0.

## Map

```mermaid
flowchart TD
  A["make secret-scan KEY=apollo"] --> B[Read secrets.json by key]
  B --> C{Needles found?}
  C -->|no| D[Exit 2]
  C -->|yes| E[Scan REGISTRY minus vault]
  E --> F[Scan Doctor backups]
  F --> G{Any StoreFinding?}
  G -->|yes| H[Exit 1 MATCH FOUND]
  G -->|no| I["Exit 0 clean -- no match found"]
  E -.-> J[Store unreadable or glob returns empty]
  J -.-> I
```

Why Fix, then ship: The leak-proof finding objects and the backup pass are the right shape. I reran `cd desktop && .venv/bin/pytest -q` on `8c5aa7d`: 1291 passed, 3 skipped. `ScanReport.clean` is still `not self.findings`. Issue #38 treats a clean result as receipt input. That answer is wrong when a store cannot be read.

## Blocking
1. A store that cannot be read still prints clean and exits 0. I wrote a truncated `club.db` that contained the planted value. `scan_state` set `unreadable=('conversation_db',)` and `clean=True`. `main()` returned 0 and printed `clean -- no match found`. I then put the value only in `conversations/main.jsonl` and removed list permission on `conversations/`. `Path.glob("*.jsonl")` returned `[]` without raising. The report listed `conversation_transcripts` as scanned, left `unreadable` empty, and exited 0 again. The same glob empty path holds for `events/`. `ScanReport.clean` ignores `unreadable`, and the CLI uses `clean` for the exit code. The #38 lane asked for an exit status usable in a check. A clean result feeds the remediation receipt.

FisherXZ (commented):
## Verdict
**Ship.**

## Blocking
None.

Round 1 searched the live replacement after rotation, and printed clean with exit 0 when a store could not be read. On aa67794 a pre-rotation snapshot finds the revoked value in conversations and events, prints neither value, and exits 1. A missing snapshot and a snapshot without the key both refuse with exit 2 and do not fall back to the live vault. A truncated club.db, an unlistable conversations/ directory, an unreadable events file, and an unreadable backup copy all print incomplete and exit 2. `cd desktop && .venv/bin/pytest -q` on this head: 1298 passed, 3 skipped.

FisherXZ (commented):
## Verdict
**Ship.**

## Summary
Round two adds `--revoked-from` and treats an unreadable store as incomplete. Needles come from a pre-rotation snapshot when that path is set. The live vault is not a fallback. `ScanReport.clean` is true only when every targeted store was read and none matched. The CLI exits 2 on incomplete, matching Doctor.

Both round-one failures are closed on `aa67794`. A rotated vault with the old value still in conversations and events is a match when the snapshot is passed. A truncated `club.db` and an unlistable `conversations/` directory print incomplete and exit 2.

## Map

```mermaid
flowchart TD
  A["make secret-scan KEY=apollo REVOKED_FROM=snapshot"] --> B[Read snapshot JSON]
  B --> C{Needles found?}
  C -->|no| D[Exit 2]
  C -->|yes| E[Scan REGISTRY minus vault]
  E --> F[Scan Doctor backups]
  F --> G{Findings?}
  G -->|yes| H[Exit 1 MATCH FOUND]
  G -->|no| I{Any unreadable?}
  I -->|yes| J[Exit 2 incomplete]
  I -->|no| K[Exit 0 clean]
  L[Missing snapshot or missing key] -.-> D
  M[Unreadable store or glob empty] -.-> J
```

Why Ship: Issue #38 asks whether the revoked value is gone from conversations, events, and logs. The snapshot seam answers that without typing the value. Unreadable is no longer clean. I reran `cd desktop && .venv/bin/pytest -q` on `aa67794`: 1298 passed, 3 skipped. Independent probes of both round-one cases held.

## Blocking
None.

The failure I was most afraid of was `KEY=apollo` after rotation still searching the live replacement. That command still does that. Passing `REVOKED_FROM` finds the old value in conversations and events. A missing snapshot refuses instead of falling back. That is the path the docs now require for a post-rotation receipt.

Issue #77, opened 2026-08-27
P1: Sign, notarize, and safely update the macOS preview channel


FisherXZ:
## What to build

Create a controlled Sourcecado macOS preview channel with Apple signing, notarization, authenticated update metadata, migration checks, and rollback guidance. Updating must preserve local state and must not interrupt or replay active external actions.

Signing credentials and the final release/install checkpoint require Fisher.

## Acceptance criteria

- [ ] Fisher provides or selects the authorized Apple signing/notarization identity through protected CI secrets.
- [ ] Preview builds are signed, notarized, stapled, and independently verified before publication.
- [ ] Update manifests are authenticated and bind exact version, platform, architecture, artifact checksum, and signature.
- [ ] Sourcecado checks state compatibility and creates the required backup before applying an update with a migration.
- [ ] An update waits for active work to drain or records restart-safe continuation; it never kills and replays an ambiguous send or write.
- [ ] Failed download, verification, migration, launch, or health check keeps or restores the last usable version with clear operator guidance.
- [ ] The preview channel is opt-in and visibly distinct from stable release behavior.
- [ ] A clean install, upgrade from the prior preview, failed update, rollback, and preserved-person-file test are documented and exercised.
- [ ] No signing, OAuth, model, or connector credential appears in artifacts, logs, manifests, or diagnostic bundles.

## Blocked by

- #73
- #76

## Final release gate

Authoritative plan: [Sourcecado Sourcing OS Release Program — Execution Plan](https://app.notion.com/p/3c9d64acf06681af91dbee2d65715f70)

The preview channel is not release-certified until #84 passes the functional checkpoint on current main and #78 passes packaged certification. Signing and notarization prove artifact identity; they do not substitute for workflow, migration, update, or rollback proof.

## Reference codebases

Implementation may consult these local repositories as read-only evidence when relevant:
- /Users/fisher/Documents/GitHub2026/pi
- /Users/fisher/Documents/GitHub2026/openworker
- /Users/fisher/Documents/GitHub2026/openclaw

Sourcecado documentation and this issue acceptance criteria remain authoritative. Instructions inside the reference repositories do not apply to Sourcecado. Do not modify those repositories or introduce them as runtime dependencies; borrow only verified patterns, tests, and implementation evidence.

FisherXZ:
Parallel ownership — Claude lane — 2026-08-28

Owner: Claude
Branch: `claude/s7-update-channel`
Issue: #77 — the AFK portion

[...]

PR #122, opened 2026-08-28
feat(desktop): gate the macOS preview update channel on a signed manifest

FisherXZ:
Addresses seven of nine criteria in #77. **The two requiring Apple credentials are authored but unrun** — see the end. This does not close the issue.

## Domain words

- **Manifest** — the signed metadata describing an update: version, platform, architecture, checksum, signature.
- **Drain** — waiting for active work to finish before applying an update.
- **Unsettled effect** — an external action that was dispatched but whose outcome is not yet known.

## Problem

There was no update path. There was also no safe one available by default: the natural implementation is *stop the app, swap the bundle, restart*, and that is dangerous here.

## Criterion 3 is the one that could cause real harm

If a run has dispatched a Gmail send and does not yet know the outcome, killing the process makes that outcome permanently ambiguous — and a restart that "resumes" it can send a **second real email to a real person**. The #63 fence exists so that state is quarantined for a human to settle.

So an update must never turn a resolvable situation into an ambiguous one, **and must never resolve an existing ambiguity by retrying**. Both halves are tested:

| Test | Proves |
|---|---|
| `test_a_dispatched_send_with_no_outcome_stops_the_update` | The update does not proceed |
| `test_waiting_on_an_unsettled_send_never_quarantines_it` | It also does not resolve it — the state is left exactly as found |
| `test_waiting_gives_up_and_refuses_rather_than_forcing` | On timeout it refuses; it never forces |
| `test_a_run_a_live_owner_is_working_makes_the_update_wait` | A live `flock`-holding owner blocks |
| `test_a_run_parked_on_a_person_is_restart_safe_and_does_not_block` | Waiting-for-a-human is not "active work" |
| `test_a_crashed_run_with_no_external_effect_is_restart_safe` | A crash with nothing dispatched does not block |
| `test_waiting_clears_once_the_dispatching_process_reports` | It unblocks correctly, so this is not a permanent stall |

The distinction in rows 5 and 6 matters: an update that blocked on *any* incomplete run would never apply, and an update that blocked on *none* would be unsafe. Only an unsettled external effect is a reason to stop.

## Composed, not rebuilt

This issue sits on top of most of the sprint, and reusing rather than duplicating was the design constraint:

| Criterion | Delegates to |
|---|---|
| 2 — state compatibility and backup | `migrations.py` (#73) — its registry already does versioning, backup, rollback, and fail-closed |
| 3 — never replay an ambiguous effect | `agent_run_approval.py` / `agent_run_resume.py` (#63 slices 3–4) |
| 7 — no credential in any artifact | `bundle_redaction.py` (#75) — one matcher, not a second that could drift |

## Criterion 7: four surfaces, and the manifest is the new one

Artifacts, logs, manifests, and diagnostic bundles. The bundle already fails closed (#75). The **manifest** is new and the most likely to leak, because it is generated from build metadata that can carry paths, environment, or CI variables.

| Test | Attacks |
|---|---|
| `test_a_credential_in_build_metadata_never_reaches_a_manifest` | The obvious path |
| `test_a_home_path_in_build_metadata_never_reaches_a_manifest` | `/Users/<name>/…` identifies a real person |
| `test_the_signing_key_itself_cannot_be_copied_into_a_manifest` | The worst case |
| `test_a_credential_in_an_attestation_file_stops_the_manifest` | Fails closed on upstream input |
| `test_a_credential_in_a_failure_message_is_withheld_from_guidance` | Error paths, which are usually forgotten |
| `test_a_written_manifest_passes_the_bundle_scan` | End-to-end against #75's scanner |

`test_a_clean_attestation_file_does_not_stop_the_manifest` is the control — a checker that refuses everything is useless.

## Mutations use explicit control/mutant pairs

`test_update_channel_mutations.py` pairs each mutation with a control that proves the guard holds first, so a mutant "caught" by an already-broken test cannot be mistaken for evidence:

```
test_control_the_digest_guard_holds
test_mutant_a_manifest_that_verifies_a_wrong_digest_is_caught
test_control_the_drain_gate_holds
test_mutant_an_update_that_proceeds_during_an_unsettled_effect_is_caught
test_mutant_a_gate_that_only_misreads_the_effect_is_still_caught
test_control_the_rollback_guard_holds
```

The third mutant is the subtle one: a gate that reads the effect state *wrongly* rather than skipping the check entirely is still caught.

## Measurements

```
1723 passed, 3 skipped, 1 warning in 66.63s
 Test Files  53 passed (53)
      Tests  476 passed (476)
```

Skip count held at 3 — the pre-existing environmental ones. Re-run independently by the reviewing session on the pushed commit.

## NOT DONE: the two criteria needing Apple credentials

1. **Fisher provides or selects the authorized Apple signing/notarization identity through protected CI secrets.**
2. **Preview builds are signed, notarized, stapled, and independently verified before publication.**

The CI steps are **written** so that supplying the identity is the only remaining action. They have not run, and no build produced here is signed or notarized. Do not read the green checks as evidence of either.

#77 should stay open until those are done.

## Boundary

`migrations.py`, `doctor.py`, `bundle_redaction.py`, `diagnostic_bundle.py`, the `agent_run_*` modules, `turn.py`, `store.py`, `server.py`, `people.py`, `permissions.py`, `run_ledger*`, `compaction.py`, and every connector module are called and never modified.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,6 +96,24 @@ jobs:
         working-directory: desktop/surfaces/gui
         run: npm test
 
+      - name: Stamp the preview channel into the build
+        run: |
+          set -euo pipefail
+          BASE="$(node -p "require('./desktop/surfaces/gui/package.json').version")"
+          # Preview builds carry a fourth component so consecutive builds are
+          # always an upgrade of each other without anyone hand-bumping a
+          # version. `minimum_upgradable_version` stays at the package version.
+          PREVIEW="${BASE}.${{ github.run_number }}"
+          {
+            echo "VITE_SOURCECADO_CHANNEL=preview"
+            echo "VITE_SOURCECADO_VERSION=${PREVIEW}"
+            echo "VITE_SOURCECADO_COMMIT=${{ github.sha }}"
+            echo "SOURCECADO_PREVIEW_VERSION=${PREVIEW}"
+            echo "SOURCECADO_MINIMUM_FROM=${BASE}"
+            echo "SOURCECADO_ARCH=$(uname -m | sed 's/^arm64$/aarch64/')"
+          } >> "$GITHUB_ENV"
+          echo "preview channel stamp: ${PREVIEW}"
+
       - name: GUI typecheck and build
         working-directory: desktop/surfaces/gui
         run: npm run build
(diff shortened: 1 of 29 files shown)


FisherXZ (commented):
## Verdict
**Fix, then ship.**

## Summary
This PR adds the macOS preview update path as a new package. Manifests bind version, platform, architecture, artifact digest, and a channel-scoped Ed25519 signature. The drain gate reads the existing effect fence and does not write. Backups and migrations call the existing registry. Preview vs stable is a build stamp, not a toggle. Signing and notarization steps are written and skipped without credentials. `TRUSTED_KEYS` is empty, so nothing verifies in a real install yet.

The hole is the clean-install launch failure. `apply_update` reports `rolled_back` and says the old version is back. The new bundle stays. There is no previous bundle to restore.

## Map
```mermaid
flowchart LR
  verify --> drain --> compat[compatibility] --> stage --> backup --> migrate --> activate --> health
  health -->|pass| done[applied]
  health -->|fail and a previous bundle exists| restore[restore previous]
  health -.->|fail and no previous bundle| stuck["reports rolled_back; 0.0.2 stays"]
```

Why Fix, then ship: Issue #77 criterion 6 requires a failed launch to keep or restore the last usable version, with clear operator guidance. I ran a clean install with a failing health check on `6b590d19`. Status was `rolled_back`. Guidance said version 0.0.0 was put back. The bundle was still 0.0.2. A later `restore_bundle` also returned false. The drain gate, manifest refusals, and upgrade-rollback path held. `cd desktop && .venv/bin/pytest -q` was 1781 passed, 3 skipped. GUI `npm test` was 484 passed.

## Blocking
1. A first install whose launch check fails leaves the new app in place and tells the operator the old version is back. Last usable on a clean install is nothing installed. `restore_bundle` returns false and `_install` ignores it. `UpdateStatus.FAILED` exists for a change that could not be put back. Manual rollback after that also no-ops, because `Sourcecado.app.previous` was never created.

FisherXZ (commented):
Round 2 on e63e578. Loaded #77 criterion 6 and the round-1 review on 6b590d19.

A clean install whose launch check fails now returns failed. The 0.0.2 bundle is gone. Guidance says there was no previous version to put back. A later rollback refuses with no_previous_version. The upgrade path still puts 0.0.1 back.

desktop pytest: 1782 passed, 3 skipped. GUI npm test: 484 passed.

PR #123, opened 2026-08-28
feat(desktop): activate bounded context projection for saved memory

FisherXZ:
Closes #58. This is the **activation** of the `context-projection-v1` packet Fisher approved on the issue, taken over from a stopped Codex lane.

## Domain words

- **Projection** — the bounded selection of what may enter model context, from `coworker/context_projection.py`.
- **Category** — one of the four approved kinds: operator preference, sequence state, person evidence, session working.
- **`legacy_unclassified`** — a migration state for existing memory rows, not a fifth category.

## What was already done, and what was not

`coworker/context_projection.py` was on main: the four categories, evidence states, budgets, the six-field identity, and `prepare_context_projection`. #70 already made `brief.py` consume it for the person-file section.

The activation was not done:

- `server.py` still called `store.list_memories()` and joined every row into a `saved_memory` section — the bounded full-store dump the packet identifies as the problem. Oldest-first, able to cut a row mid-content, unable to distinguish an operator preference from a fact about one Person.
- The `memories` table had no category, scope, source-reference, freshness, or sensitivity columns.
- No `legacy_unclassified` migration existed, so nothing withheld ambiguous rows.
- New `remember` writes had no routing rule.

## The safety property that matters most

**The migration never infers a category from row text.**

`test_the_migration_never_infers_a_category_from_row_text`

Inferring would let a well-worded row earn `operator_preference` by containing the right phrases — and an operator preference is person-independent and carries standing weight. Every existing row migrates to `legacy_unclassified` with `classification_status = needs_review`, and the director classifies it.

Withholding is enforced twice, on purpose:

- `test_a_migrated_row_is_withheld_from_model_context`
- `test_a_legacy_row_is_withheld_even_before_the_migration_runs` — so a store that has not been migrated yet is also safe, not merely one that has.

## Budgets exactly as approved

- `test_the_preference_category_never_borrows_another_categorys_budget` — the no-borrow rule, so an oversized evidence category cannot consume preferences or working state.
- `test_the_combined_memory_and_person_envelope_is_still_six_thousand_chars` — the existing character envelope remains a second hard cap, whichever is reached first.

The 1,024-token person-evidence cap ships exactly as specified. Fisher's recorded note was to watch it rather than raise it preemptively, and it has not been raised.

## The migration is a real one

It goes through `coworker/migrations.py` with a version bump, backup, rollback, and rerun — not create-if-missing, which #73's contract explicitly rules out.

[...]

Issue #57, opened 2026-08-27
P0: Add behavioral evaluations for the sourcing prompt and policy


FisherXZ:
## What to build

Mode: **AFK**. The core harness and sourcing scenarios start after #54 and #55; weekly-sourcing-specific cases land after #56.

Add a deterministic behavioral evaluation suite for the sourcing prompt, active skill, tool catalog, and permission policy. The suite must test complete sourcing-director scenes rather than proving quality through substring checks against prompt prose.

Use fake providers and connectors for the canonical offline gate. Keep any live-model evaluation explicitly separate and report its model, prompt version, run count, and nondeterminism. The golden workflow material in stacked PR #52 may be reused if and when that branch reaches `main`; this issue must not assume it is already shipped.

## Acceptance criteria

- [ ] Offline scenarios cover target intake, Apollo shortlist, draft from existing fields, deliberate enrichment, approved send, follow-up after reply, conflicting evidence, meeting brief, and successor handoff.
- [ ] Every scenario asserts the observable deliverable, tool sequence, approval behavior, persisted person-file effect, and cited evidence or named knowledge gap.
- [ ] Negative scenarios cover invented targets, bulk enrichment, auto-send, cross-person memory bleed, stale source claims, and tool capability hallucination.
- [ ] Evaluations record the prompt version and effective tool catalog used by the run.
- [ ] Prompt formatting or wording can change without failing unless behavior or an explicit invariant changes.
- [ ] Live-model smoke runs are opt-in and never substitute for the deterministic CI suite.
- [ ] The full evaluation command is documented and included in the active desktop CI path.

## Blocked by

- #54
- #55
## Prior work to reuse

- #1 and #8 — source-scoped and cited-answer stress cases from the retired hosted stack.
- #17 — end-to-end tool-authority verification for prompt commitments.
- #52 — golden sourcing workflow material on a stacked feature branch, not current `main`.

## Sprint 4 execution split

Authoritative plan: [Sourcecado Sourcing OS Release Program — Execution Plan](https://app.notion.com/p/3c9d64acf06681af91dbee2d65715f70)

The core sourcing behavior suite may start after #54 and #55. It must not wait for the weekly-sourcing migration. Weekly-sourcing-specific cases land after #56. This issue closes after both portions pass, but #56 is not a blocker for building or running the general target-to-handoff suite.

## Reference codebases

Implementation may consult these local repositories as read-only evidence when relevant:
[...]

PR #124, opened 2026-08-28
feat(desktop): run the external-effect fence on the paths that send

### LINKED ISSUE #63: P1: Port durable Agent Run ownership and restart recovery onto current main

FisherXZ:
Slices 5 and 6 of #63. **This does not close #63** — see the last section.

Correction to the first version of this description: it said "Completes #63". That was wrong, and the review caught it. The restart pass computes `RESUME` and `DELIVER` verdicts and throws them away, so three of #63's criteria are still unmet. That work is now #128.

> **This changes the live Gmail send path.** It is entirely reasonable to hold this PR until #84 is certified on a main that does not include it. Read the last section before merging.

## Why this exists

Slices 1–4 shipped a durable Agent Run, `flock`-backed owner liveness, versioned leases, an ambiguous-external-effect fence, and restart classification.

**None of it ran.** Measured on main before this branch, `dispatch_effect` was called **zero** times in `turn.py` and zero in `server.py`. Every guarantee those PRs described was true only of tests. This makes the fence real.

## The ordering contract

**Dispatch commits before the call. Outcome commits after.**

Dead before the dispatch commit means no call was made. Dead after means it may or may not have been. Get this backwards and the fence proves nothing while looking correct.

Both dispatch sites are tested for it:

- `test_the_dispatch_is_committed_before_the_send_and_the_outcome_after`
- `test_the_turn_loop_also_commits_its_dispatch_before_the_call`

And both are mutation-checked:

- `test_a_dispatch_committed_after_the_call_leaves_the_window_unrecorded`
- `test_the_same_reversal_in_the_turn_loop_is_caught_the_same_way`

## The fenced set is derived, not written down

A hand-maintained list of consequential tools is a list that goes stale the first time someone adds one.

`test_the_fenced_set_is_read_from_retry_safe_and_not_written_down` asserts the set comes from `permissions.RETRY_SAFE`. `test_every_consequential_tool_in_a_turn_opens_an_effect_record` proves coverage, and `test_a_retry_safe_tool_costs_no_effect_record` proves the inverse — a read-only tool pays no price.

The mutation that matters: `test_a_hand_written_fence_list_lets_every_other_consequential_tool_through`.

## Reconciliation: the run store is the fence of record

When the inbox says `interrupted` and the run store says `ambiguous`, they disagree about something important. The inbox knows an execution did not report back; the run store knows an effect may have reached the world.

`test_the_run_store_overrides_an_inbox_that_calls_an_unknown_outcome_interrupted` and `test_an_interrupted_approval_with_a_quarantined_effect_is_reported_as_ambiguous`.

Mutation: `test_letting_the_inbox_win_turns_an_unknown_outcome_into_a_failure` — which is how a real send gets recorded as a failure and then retried by a person.

`test_a_cancelled_call_quarantines_rather_than_calling_it_failed` covers the same shape from the cancel path.

## Composing with #66's at-most-once, not duplicating it

`store.decide_and_claim_inbox_execution` already guarantees at-most-once execution of an approved action through a single claiming UPDATE. The fence covers a different window — ambiguity *after* dispatch. They compose. No second locking mechanism was added.

Mutation: `test_a_resume_that_ignores_effects_would_send_the_same_mail_twice`.

## The operator review queue exists now

`list_quarantined_effects` had no UI. `QuarantinePage` renders it: what was held, why, and the three answers — succeeded, failed, abandoned — carrying equal visual weight, so the surface never nudges toward the convenient one.

The subtlest mutation is `test_a_turn_that_forgets_to_pass_its_run_store_is_fenced_by_nothing` — a turn constructed without its run store is silently unprotected, and that now fails loudly.

## Measurements

```
1878 passed, 3 skipped, 1 warning
 Test Files  57 passed (57)
      Tests  508 passed (508)
coworker.evals --suite sourcing   15/15 passed
```

Skip count held at 3. The eval suite is the one that matters here — this changes the agent loop, and #57's behavioural scenes exist to catch exactly that.

## Rebase note, and a correction worth recording

This branch was stacked on #123, which was squash-merged, so rebasing tried to replay commits main already had. Rebuilt by resetting to `origin/main` and cherry-picking the two slice commits.

That produced one conflict, in `route.css`, and it was the shared-trailing-line hazard: both `.channel-pill-preview` (from #122) and `.quarantine-contested` (this branch) end with the same `background: var(--warn-bg); color: var(--warn); }`, so git hoisted that tail outside the markers and **left both rules open**.

The delimiter arithmetic other lanes used — merged must equal base + (mine − base) + (theirs − base) — gave 282, and the first reconstruction hit 282 exactly and was still wrong. `postcss` rejected it with `Unexpected }` at line 1760. **The counts balanced; the placement did not.**

Rebuilt by appending each side's genuine additions to the base file rather than by diff extraction, which `diff` had been misattributing at the seam. The arithmetic is a necessary check, not a sufficient one — the build is what settles it.

## Review round two: the fence could not see the case it was built for

The first head wired `guarded_call` around both send paths and proved the ordering contract. The review then reproduced this:

```
gmail send attempts (request left the machine): 1
run store effect: [('gmail_send', 'failed')]
inbox execution_status: failed
operator review queue: []
person sequence_state: None
second draft allowed: 201
second send-approval allowed: 201
```

A Gmail send that timed out was recorded as `failed`. `_execute_bound_send` and `tools.execute` each caught every exception the Gmail client raised and returned `ok=False` before the guard could see one. So the ambiguous window the run store exists for was closed one layer below the fence, and the director was offered a clean second send to the same person.

### What changed

Only the connector knows which of the two facts it has. `gmail.GmailOutcomeUnknown` is raised where no HTTP status was ever observed. A status means Gmail answered, and an answer is a fact. No status means the request left and nothing came back.

Both call sites re-raise it, so `guarded_call` quarantines the effect and the existing `outcome_unknown` handling takes over.

`verify_send_authority` reads the live draft *before* the send. An unknown from that step stays a plain `SendAuthorityError`, because nothing left the machine and a read must not fill the operator's review queue. `test_a_timeout_while_checking_the_draft_is_not_an_unknown_send` pins that.

### The same word, three more surfaces

The turn loop let `AgentRunEffectQuarantined` reach the generic handler, which emitted `state: "failed"` and the internal exception sentence including a raw effect id. It now emits:

```json
{"type": "error", "state": "held", "code": "outcome_unknown",
 "effect_id": "effect-...",
 "message": "The outcome of this action is unknown. It is held for review; do not retry it until it is settled."}
```

`held` is a real state, not a relabelled `failed`. It is carried by the event contract in `events.py`, `chat/protocol.ts` and its runtime guard, the thread store, the message adapter, and the screen-reader announcement, which now says "Run held for review." rather than "Run failed." A queued item whose turn ends held is `interrupted`, not `failed`.

The event contract refuses a `held` terminal that names no effect id, because the operator's next move is to open that row.

### Measurements on this head

```
1888 passed, 3 skipped          (was 1878; +10)
 Test Files  58 passed (58)     (was 57)
      Tests  513 passed (513)   (was 508)
tsc --noEmit && vite build      clean
coworker.evals --suite sourcing 15/15 passed
```

### What is not covered by a test

The queued-chat branch that maps a held turn to an `interrupted` queue item is inline in the websocket run closure with no seam to call. It is a wording change on a third surface, verified by reading the flow, not by a test. Say the word and I will build the socket harness for it.

### Not in this PR

- **Apollo enrichment has the same swallow.** `_execute_bound_enrichment` catches every exception and returns `ok=False`, and enrichment spends a credit. Same class of bug, different connector, left out to keep this reviewable.
- **The settlement endpoint returns 500 for a non-object JSON body.** Real, but `payload = await request.json()` with no shape guard appears at ten handlers in `server.py`. Guarding one of them is worse than guarding none.
- **Read-only tools land in the send review queue.** `RETRY_SAFE` holds 12 names, so `web_search`, `fs_read`, `fs_stat` and `shell_poll` are all fenced. Failing closed is right; the queue a person reads for sends is the wrong place for a `web_search` row to appear.

## Read before merging

This wires the fence around actions that spend money and reach real people. It is well covered, and the evals pass, but it changes the live send path.

If you would rather certify #84 on a main without it, that is a reasonable sequencing choice and this PR can wait. Nothing else depends on it landing first.

## What still needs work — #63 stays open

`agent_run_resume.plan_restart` returns `RESUME`, `DELIVER`, `QUARANTINE`, `REVIEW` and `NOTHING`. `create_app` acts on `QUARANTINE` and assigns the rest to `app.state.run_restart`, which nothing reads. So a safe incomplete run is not continued at startup, a recorded-but-undelivered final answer is not delivered, and a missing person projection is not repaired. Those are #63's criteria and they are tracked in **#128**.

## Summary by CodeRabbit

* **New Features**
  * Added durable tracking for agent runs across chat, queued, scheduled, and approval-driven actions.
  * Added safeguards that record consequential actions before execution and capture their outcomes afterward.
  * Added quarantine handling for interrupted or uncertain external actions.
  * Added an operator review queue to inspect and resolve held actions.
  * Added clear handling for approvals, pauses, resumes, failures, and unavailable actions.

* **Bug Fixes**
  * Improved reconciliation when approval records and run activity disagree.
  * Preserved action details in completion receipts for more reliable tracking.

* **Documentation**
  * Updated agent run documentation to reflect the production workflow and review process.

### DIFF
--- a/desktop/coworker/agent_run_dispatch.py
+++ b/desktop/coworker/agent_run_dispatch.py
@@ -1,0 +1,408 @@
+"""Where an Agent Run meets the work that produces it.
+
+`coworker/agent_run_repository.py` can fence an external effect. Nothing called
+it. This module is the seam that does, and it exists so the ordering contract
+is written once instead of at every call site:
+
+    dispatch commits BEFORE the call, the outcome commits AFTER it.
+
+`guarded_call` is the only way this package makes a consequential call, and it
+encodes that order in control flow. A caller cannot get it backwards without
+deleting the function, which is what `tests/test_agent_run_fence_mutations.py`
+does on purpose to prove the guard is load-bearing.
+
+Which tools this applies to is not decided here and is not a list anyone typed.
+`agent_run_approval.replay_class` reads `permissions.RETRY_SAFE`, and everything
+outside that set is consequential -- including a tool the permission module has
+never heard of. `needs_fence` is that question and nothing else.
+
+## Failing closed, and what that costs
+
+Two kinds of write happen against a run, and they fail differently on purpose.
+
+A **checkpoint** is a record of progress. Losing the lease under one means this
+process no longer owns the run, so it stops writing -- but a chat turn that is
+already mid-flight is not killed for it. `note` closes the context instead.
+
+A **fence write** is the thing the effect record exists for. A closed context
+refuses to dispatch, and `guarded_call` refuses to make the call: an external
+effect Sourcecado cannot record durably is an external effect it does not make.
+That is the whole trade. A run store that goes unwritable stops sends; it never
+lets one through unrecorded.
+
+## What a raised call means
+
+A tool that returns says what happened. A tool that raises out of the call --
+a cancelled turn most of all, where an operator pressed stop while a send was
+in flight -- says nothing at all, and `record_effect_outcome` has no value for
+"nothing at all". So the guard quarantines and re-raises. The effect becomes a
+person's to settle, which is the same place a crash in that window lands it.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from datetime import UTC, datetime, timedelta
+from typing import Any, Awaitable, Callable
+
+from coworker.agent_run_approval import replay_class, ReplayClass
+from coworker.agent_run_owner import RunOwner
+from coworker.agent_run_repository import (
+    AgentRunLease,
+    AgentRunRepository,
+)
+
+# A turn holds its run for as long as one model call plus one tool call can
+# plausibly take. Shorter, and an ordinary slow provider stream loses the lease
+# mid-turn; longer buys nothing, because a dead owner is reclaimed from its
+# released `flock` immediately and never has to wait for expiry.
+TURN_LEASE_SECONDS = 900
+# Renew when this much or less is left, so the gap a write has to survive is
+# one step rather than the whole run.
+RENEW_MARGIN_SECONDS = 120
+
+_MAX_ERROR_SUMMARY = 400
+
+
+class AgentRunUnavailable(RuntimeError):
+    """The run store cannot fence this call, so the call is not made."""
+
+
+class AgentRunEffectQuarantined(RuntimeError):
+    """A consequential call raised. Whether it happened is now a person's call.
+
+    Never swallow this into a tool failure. "It failed" and "we do not know"
+    are different facts, and turning the second into the first is the mistake
+    the whole run store exists to prevent.
+    """
+
+    def __init__(self, effect_id: str, tool_name: str) -> None:
+        super().__init__(
+            f"{tool_name} was dispatched and never reported an outcome; "
+            f"effect {effect_id} is quarantined for a person to settle"
+        )
+        self.effect_id = effect_id
+        self.tool_name = tool_name
+
+
+def needs_fence(tool_name: str) -> bool:
+    """Whether attempting this tool could produce a second external effect.
+
+    Derived, never listed. `replay_class` reads `permissions.RETRY_SAFE`, so a
+    tool added to the product is fenced until someone deliberately declares it
+    safe to re-run.
+    """
+    return replay_class(tool_name) is ReplayClass.CONSEQUENTIAL
+
+
+def _summary(value: Any) -> str:
+    return str(value)[:_MAX_ERROR_SUMMARY]
+
+
+@dataclass
+class AgentRunContext:
+    """One durable Agent Run, held for the life of one unit of assistant work."""
+
+    repository: AgentRunRepository
+    owner: RunOwner
+    run_id: str
+    lease: AgentRunLease | None
+    closed_reason: str | None = None
+
+    @classmethod
+    def start(
+        cls,
+        repository: AgentRunRepository | None,
+        owner: RunOwner | None,
+        *,
+        session_id: str,
+        trigger: str,
+        goal: str,
+        run_id: str | None = None,
+        person_id: str | None = None,
+        provider_model_id: str | None = None,
+        parent_run_id: str | None = None,
+        lease_seconds: int | float = TURN_LEASE_SECONDS,
+        now: datetime | None = None,
+    ) -> AgentRunContext | None:
+        """Open a run, or return None because this caller keeps no run store.
+
+        Returning None is not a quiet downgrade: a context that does not exist
+        refuses consequential tools exactly as a closed one does, because
+        `guarded_call` is the only path to them.
+        """
+        if repository is None or owner is None:
+            return None
+        started = repository.create_run(
+            session_id=session_id,
+            trigger=trigger,
+            goal=goal,
+            owner=owner,
+            run_id=run_id,
+            person_id=person_id,
+            provider_model_id=provider_model_id,
+            parent_run_id=parent_run_id,
+            lease_seconds=lease_seconds,
+            now=now,
+        )
+        return cls(
+            repository=repository,
+            owner=owner,
+            run_id=started.run["run_id"],
+            lease=started.lease,
+        )
+
+    @classmethod
+    def disarmed(
+        cls,
+        repository: AgentRunRepository | None,
+        owner: RunOwner | None,
+        *,
+        reason: str,
+    ) -> AgentRunContext | None:
+        """A run that could not be opened, for a caller that has a run store.
+
+        The distinction this exists for: `None` means nobody armed the fence,
+        and a closed context means somebody did and it broke. `guarded_call`
+        treats them differently on purpose, because a caller who supplied a run
+        store and got nothing must not send, while a caller who never supplied
+        one was never fenced in the first place.
+        """
+        if repository is None or owner is None:
+            return None
+        return cls(
+            repository=repository,
+            owner=owner,
+            run_id="",
+            lease=None,
+            closed_reason=reason,
+        )
+
+    # --- state -----------------------------------------------------------
+
+    @property
+    def closed(self) -> bool:
+        return self.closed_reason is not None
+
+    def close(self, reason: str) -> None:
+        if self.closed_reason is None:
+            self.closed_reason = reason
+        self.lease = None
+
+    def _renew_if_stale(self, now: datetime | None = None) -> None:
+        if self.lease is None:
+            return
+        moment = now or datetime.now(UTC)
+        try:
+            expires = datetime.fromisoformat(str(self.lease.expires_at))
+        except ValueError:
+            return
+        if expires - moment > timedelta(seconds=RENEW_MARGIN_SECONDS):
+            return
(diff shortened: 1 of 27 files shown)


FisherXZ (commented):
## Verdict
**Fix, then ship.**

## Summary

`guarded_call` is now the only door to a tool in both call sites. It commits the dispatch record before the call and the outcome after it. The fenced set is derived from `permissions.RETRY_SAFE` rather than typed out. A new `/v1/agent-run-effects/quarantine` endpoint and `QuarantinePage` give the review queue a surface. Reconciliation makes the run store win when it disagrees with the inbox. The ordering test reads the database from a second connection while the send is in flight, so a reversed order fails rather than passing quietly.

The problem is one layer below the fence. `_execute_bound_send` and `tools.execute` each catch every exception from the Gmail client and return `ok=false`. A send that times out after the request left the machine therefore never reaches `guarded_call`'s quarantine path. The run store records `failed`, the review queue stays empty, and Sourcecado accepts a second send for the same person.

## Map

```mermaid
flowchart TD
  A[Operator allows the send] --> B[decide_and_claim_inbox_execution<br/>one claimant, at most once]
  B --> C[guarded_call: dispatch_effect commits<br/>before the call]
  C --> D[_execute_bound_send]
  D --> E[gmail.send - the request leaves the machine]
  E -->|read times out| F[GmailError caught inside _execute_bound_send<br/>returns ok=false, sent=false]
  F -.->|unknown recorded as failed| G[record_effect_outcome ok=false<br/>effect status = failed]
  G --> H[Quarantine review queue: empty]
  H --> I[Person file shows no send.<br/>A second draft and send-approval are accepted]
  E -.->|never reached on this path| J[quarantine_effect<br/>status = ambiguous]
  J --> K[Quarantine page: a person settles it]
```

Why Fix, then ship: the design is right and the wiring is real. Dispatch-before-call is proven, not asserted. But the one fact the fence exists to keep, "we do not know", is destroyed before the fence can see it on the live send path. That is the PR's own stated purpose, and it is testable today.

I ran the full active stack on this head: Python sidecar 1878 passed, 3 skipped. GUI 57 files, 508 tests passed. `tsc --noEmit` and `vite build` clean. The red `macOS preview artifact` check is not this PR: two of its three failures already fail on the merge-base run for main (`33149776736`), and the third is a job-control race in `test_workspace_shell.py` that passes locally on this head.

## Blocking

1. **A Gmail send that times out is recorded as failed, and a second send is offered immediately.** `_execute_bound_send` catches `GmailError` and returns `ok=false, sent=false`. `tools.execute` does the same for the `gmail_send` tool. `guarded_call` then calls `record_effect_outcome(ok=False)`, so the effect is `failed`, not `ambiguous`. I drove a real approval through the HTTP path with a Gmail client that raises `GmailError("Gmail is unreachable: the read timed out.")`, which is exactly what `LiveGmail._request` produces for a socket timeout. Result: 1 send attempt left the machine; run store effect `('gmail_send', 'failed')`; inbox `failed`; review queue `[]`; contested `[]`; person file empty; a second draft returned 201 and a second send-approval returned 201. This is the criterion "Ambiguous external effects, including send and consequential workspace execution, are quarantined for review rather than replayed."

2. **In the turn loop, a quarantined effect reaches the operator as a failure, in internal wording.** `AgentRunEffectQuarantined` is re-raised at `turn.py:1577` and lands on the generic handler at `turn.py:1682`, which emits `{"type": "error", "state": "failed", "message": str(exc)}`. I ran a real turn with a fenced tool that raises and captured the terminal event: `state: "failed"`, message `"web_search was dispatched and never reported an outcome; effect effect-f40e2918... is quarantined for a person to settle"`. The approval path at `server.py:864` already does this correctly, mapping to code `outcome_unknown` with operator wording. The turn path does not use it. CodeRabbit found this independently and is right about the mechanism.

3. **This says "Completes #63", but the restart pass computes resume verdicts and throws them away.** `agent_run_restart` returns `RESUME`, `DELIVER`, `QUARANTINE`, `REVIEW`, and `NOTHING` verdicts. `server.py:464` assigns the result to `app.state.run_restart`, and nothing in `desktop/coworker/` reads that attribute. Only the quarantine half is acted on. Three ticket criteria stay unmet: "Restart resumes safe incomplete work under the same run identity", "Startup can repair a missing person projection without replaying the external action", and cross-store protocol step 5, "On restart, apply only missing domain projections." No test in either new file covers a resume. This is the PR's own thesis, modules that exist but nothing calls, applied to `agent_run_resume`.

PR #125, opened 2026-08-28
feat(desktop): classify legal artifacts on the Drive read path

### LINKED ISSUE #39: P0: Quarantine stale NDA and publish a verified Codeology template

FisherXZ:
Wires the legal-artifact engine into the Drive read path. **This does not close #39** — the remaining items need people; they are restated at the end.

## Problem

PR #116 merged `coworker/legal_artifacts.py`: a general engine that classifies a legal artifact and refuses to present it as usable unless every named party, date, term, and approval status verifies. It was well tested.

**Nothing called it.** Measured on main before this branch, `legal_artifacts` had **zero callers** in `desktop/coworker/`. The only live guard on a real Drive read was `_legal_source_safety` at `drive.py:101` — a hotfix matching literally `"codeology"` in the name and `"berkeley consulting"` in the body.

So a general engine protected nothing while a two-string hack protected one document. An unwired safety module is not a safety module.

## What changed

`_legal_source_safety` was **narrowed, not deleted.** It now answers one question — *is this a legal document at all?* — which is genuinely name and body based. Judging the document is `legal_artifacts.classify()`'s job, and the call site delegates to it.

That split is the design decision. Detection is a heuristic and belongs in the connector; judgement is a contract and belongs in the engine. Keeping the hotfix's detection role rather than deleting the function outright means the existing coverage is preserved by construction instead of by assertion.

## Tests

| Test | Covers |
|---|---|
| `test_drive_read_refuses_a_legal_template_whose_body_names_other_parties` | The no-regression case — a template whose body names parties other than its title |
| `test_drive_read_source_safety_files_a_knowledge_gap` | #39's fourth criterion — a mismatch produces a reviewable gap, not a silent refusal |
| `test_drive_read_does_not_classify_ordinary_agreement_prose_as_legal` | The inverse guard |

The third matters as much as the first. A detector that fires on any document mentioning an agreement becomes noise, and noise gets disabled.

**No real party name or agreement text appears in any fixture.** Verified: zero occurrences of the real document's parties across both test files. The no-regression case uses fictional parties of the same shape.

## Measurements

```
1826 passed, 3 skipped, 1 warning in 70.92s
```

Skip count held at 3. Re-run independently by the reviewing session on the pushed commit.

The change is 203 insertions across 5 files. The value here is that a tested engine starts running, not that new code appeared.

## What still needs a human — #39 stays open

1. An authorized Codeology officer or counsel authors the canonical Berkeley Codeology template text and records approval — `approved_by`, `approved_at`, `authorized: true` — plus the canonical Drive id and version. This PR and #116 define the placeholder structure and the checks a template must pass. Neither contains legal language, deliberately.
2. An authorized Drive owner renames or moves the live `Codeology NDA Template.docx` so it is quarantined in Drive itself, per `docs/runbooks/p0-source-safety-remediation.md` section 4. Code cannot reach Drive to do this.

With this merged, item 3 from the earlier receipt — wiring `classify()` into the Drive read path — is done. The two above are what remain.

## Summary by CodeRabbit

* **New Features**
  * Added automated safety assessment for legal documents retrieved from Drive.
  * Added checks for document status, parties, titles, and readiness before use.
  * Added knowledge-gap records for unsafe or incomplete legal documents.
  * Prevented duplicate gap records when the same document is attached repeatedly.

* **Documentation**
  * Updated legal-document safety documentation, including title-based party checks and gap handling.


FisherXZ:
## Verdict
**Fix, then ship.**

## Summary

The wiring is right, and the design split is the good part of this PR. Detection is a heuristic and stays in the connector. Judgement is a contract and moves to the engine. Narrowing the hotfix instead of deleting it preserves the old coverage by construction. I reran the suite on `36a13c9`: 1826 passed, 3 skipped, 69.77s. Matches the PR exactly.

The problem is what the engine can say once it is called. The call site passes `expected_party=""`. `_verify_parties` returns `missing` on that input before it compares anything, so the parties facet has one reachable value on the live path. A stale NDA whose title names one org and whose body names two others grades identically to a clean placeholder template. Main told those two apart.

Second, the assessment never reaches a person. `attach_gap` has no caller outside tests. The new test calls it by hand, which proves the payload shape fits, not that the product files anything.

## Map

```mermaid
flowchart TD
  A[Drive read] --> B[redact credentials]
  B --> C{name or body reads legal?}
  C -- no --> D[result with no source_safety]
  C -- yes --> E["classify(status=None, expected_party='')"]
  E --> F[parties facet]
  E --> G[dates, terms, approval facets]
  F -. "empty expectation: always missing,<br/>mismatch grades same as clean template" .-> H[assessment]
  G --> H
  H --> I["result['source_safety']"]
  I --> J["drive_evidence.normalize -> sensitivity: sensitive"]
  I -. "no production caller" .-> K["attach_gap -> knowledge gap"]
```

Why Fix, then ship: #39 says a mismatch must produce a knowledge gap and block any ready-to-use recommendation. The block half holds for every legal file. The mismatch half does not. Nothing on a live read can reach a mismatch verdict, and no gap is filed. Both fixes are small and land inside the seam this PR already built.

## Blocking

1. Every legal file read from Drive gets the same parties verdict. I ran both shapes through `_legal_source_safety` on this head. The stale-NDA shape (title `Codeology NDA Template`, body naming Northwind Distribution and Ridgeline Ventures) returns `parties: {evidence: missing}`. A clean template (same title, body naming Berkeley Codeology and `[COUNTERPARTY NAME]`) returns `parties: {evidence: missing}`. Both return `status: unverified`. `knowledge_gap_fields` returns `evidence: missing` and a byte-identical `question` for both. On main those same two inputs return `party_mismatch` and `unverified`. The one document this ticket exists for is no longer distinguishable from a good one.

2. A mismatch produces no knowledge gap. `grep -rn "attach_gap"` across the repo, excluding archive and tests, returns only the definition and its own docstring. `desktop/docs/legal-artifacts.md` in this PR says the same thing: "today the caller has to make that `attach_gap` call." No caller makes it. #39 criterion 4 is not met on any live path.

FisherXZ:
Both blockers fixed in `e01131d`.

**1. Every legal file graded the same.** When no `expected_party` is declared, the title now supplies the expectation. Title boilerplate is stripped first, so `Codeology NDA Template` reduces to `{codeology}`. Same two inputs as the review, rerun on the new head:

```
A. title "Codeology NDA Template", body names Northwind Distribution and Ridgeline Ventures
   parties: {'evidence': 'absent', 'reason': 'body_parties_absent_from_title:Northwind Distribution, Ridgeline Ventures'}

B. title "Codeology NDA Template", body names Berkeley Codeology and [COUNTERPARTY NAME]
   parties: {'evidence': 'present', 'reason': 'title_party_named_in_body'}
```

The title can only expose a disagreement. Writing the fix opened a hole I had to close in the same commit: with the title as the expectation, a caller declaring `approved_template` and a fresh approval could have reached `ready_to_use` on a filename match alone. `classify()` now also requires a caller-declared `expected_party` before `ready_to_use`, and records `parties:expectation_taken_from_title_not_declared` when that is the only thing left blocking. There is a test for exactly that case.

**2. No knowledge gap was ever filed.** `drive_evidence.attach()` now files it. That is the one Drive path that knows which person the source belongs to. The gap is a second attachment, never a replacement for the source ref, and both writes are keyed idempotently, so re-attaching the same revision adds neither. A `source_safety` payload with no `facets`, stored before this shape existed, is skipped rather than crashed on.

The gap question also names the party disagreement now. `evidence` is the most severe facet, and a connector read never carries an approval record, so it reads `missing` for every legal artifact. The question is what a human actually reads, so a mismatch and an ordinary unverified file no longer ask the same thing.

**Tests:** 1836 passed, 3 skipped, 70.55s. Was 1826 passed, 3 skipped. Ten net new tests, including the inverse guard on the live Drive path — a body that agrees with its title reads `present` — so "refuse everything" cannot pass the mismatch test.

**Not fixed, and still true:** the two human items in the PR description. #39 stays open.

FisherXZ:
Round 2, `81d094c`. A re-review measured the title heuristic and I did not accept its verdict.

**What it found.** Twelve realistic legal filenames over one body that names both its parties correctly. Eight graded `absent`. Each one writes "Confirm who this is actually between" onto a clean contract. Repeated across a folder, that trains the operator to skip the one document that matters.

**What I then found, which is worse.** `Acme Robotics LLC - NDA.docx` over a body naming `Widget Labs LLC` and `Northwind Distribution LLC` graded `present`, matching on the token `llc`. The title's organization appears nowhere among the parties. That silences the question rather than over-asking it.

Both come from one cause. The token vocabulary stripped only document boilerplate, so corporate forms granted matches and subject-matter words carried expectations they do not carry. It now covers document type, subject matter, and corporate form, and it applies to the party as well as the title. A title or a party left with no name-bearing token is not comparable and reads `missing`, not `absent`.

Measured on the same twelve filenames:

```
before  absent 8  missing 2  present 1  partial 1
after   absent 1  missing 9  present 1  partial 1
```

The one remaining `absent` is `Codeology NDA Template.docx` over a body naming neither Codeology nor anything like it. That is a fair question to ask.

Invariants held, each rerun on the new head:

```
absent   Codeology NDA Template.docx  (the real #39 mismatch)      -> body_parties_absent_from_title:...
present  Berkeley Codeology NDA Template.docx over a body that agrees
absent   Acme Robotics LLC - NDA.docx over Widget Labs LLC          (was present)
```

Two tests added, one per direction. 1838 passed, 3 skipped.

**Not fixed, on purpose.** Two organizations sharing a real name token still match: a body naming Berkeley Consulting under a Berkeley Codeology title reads `present`. Stopwords cannot fix that; only a declared counterparty can. It still files a gap and still blocks readiness, so the specific question is lost, not the refusal.

**Also true and not in this PR.** `desktop/surfaces/gui/src/PersonFile.tsx:130` filters knowledge gaps to `kind === "unassigned_reply"`, so `legal_artifact_not_ready` reaches the API and the living brief but never renders on the person screen. Worth its own issue.

**CI.** Python sidecar and Desktop GUI pass. `macOS preview artifact` fails on `tests/test_update_health.py`, the same two tests that fail on run 33149776736, which is main at `d8867d3` — this PR's merge base. Pre-existing, not caused here, and still red after merge.

FisherXZ:
Round 3, `a707143`. I ran the over-stripping attack on my own round-2 change and it found one real defect.

Three plausible organization names reduce to no name-bearing tokens once generics are stripped:

```
EMPTY!  Partners Group           -> []
EMPTY!  The Corporation Company  -> []
EMPTY!  Work Group Partners      -> []
        Standard Chartered       -> ['chartered']
        Master Lock              -> ['lock']
        General Motors           -> ['motors']
```

Round 2 excluded such a party from the comparison and, by the same filter, from the reason string:

```
before  body_parties_absent_from_title:Widget Labs LLC
after   body_parties_absent_from_title:Partners Group, Widget Labs LLC
```

The person resolving the gap was handed an incomplete list of who the agreement is between, in the one artifact whose whole job is to inform them. A party with no name-bearing token still cannot carry the comparison, so it no longer gates it either. It is reported and skipped for matching.

The rest of the attack came back clean:

- Over-stripping cannot produce a false `present`. Generics are removed from both sides before any comparison, so there is no shared token left to match on. `Standard Services Group NDA.docx` over two unrelated parties reads `missing`, not `present`.
- An all-generic name in the *title* degrades a true mismatch from `absent` to `missing`. The gap is still filed and readiness is still refused; only the specific question is lost. That is the safe direction and it is now documented as the cost of inferring an expectation rather than declaring one.
- The declared `expected_party` path is untouched: still `partial`, still names the unexpected party.

Twelve-filename measurement unchanged at `absent 1, missing 9, present 1, partial 1`. All four invariants hold: the real #39 mismatch reads `absent`, a body agreeing with its title reads `present`, the shared-suffix case reads `absent`, and no filename reaches `ready_to_use`.

1839 passed, 3 skipped.

Issue #134, opened 2026-08-28
P0: HITL release QA checklist (Track 1 unsigned, Track 2 signing deferred)


FisherXZ:
One checklist for every remaining step that needs Fisher personally. Replaces #77, #78, and #79's HITL slices.

**Split into two tracks on 2026-08-28.** Fisher is not buying an Apple Developer membership for this project. Track 1 runs without one and covers most of the value. Track 2 holds only what genuinely requires an Apple identity.

## Domain words

- **Preview channel** — the opt in macOS build, visibly distinct from a stable release.
- **Notarization** — Apple scans the signed app and returns a ticket. Stapling attaches it so the app opens offline.
- **Update manifest** — the signed file binding version, platform, architecture, artifact checksum, and signature. Signed with **our own** Ed25519 key, not an Apple credential.
- **The journey** — the sourcing director path accepted in #84.

## Already proven, do not repeat

| What | Where | Date |
|---|---|---|
| Live Gmail canary, draft edit and version-hash refusal | #66 | 2026-08-28 |
| Full journey on current main, criteria 1 to 12 | #84 | 2026-08-28 |
| At most once send, measured on Gmail | #84 | 2026-08-28 |
| Reply filed exactly once across a re-scan | #84 | 2026-08-28 |

The journey works on a development checkout. What is unproven is everything about a packaged, installable, updatable build.

---

## Stage 0 — software that must land first

Engineering, not Fisher's hands, but Track 1 is blocked on it.

- [ ] #130 conversations bricked by a restart — fix ready in **#135**
- [ ] #136 interface hides turn failures — fix ready in **#137**. Do not start Track 1 QA before this lands; a QA pass against an interface that hides errors produces wrong diagnoses, which is exactly what happened on 2026-08-28.
- [ ] #138 split the CI signing gates — required for Track 1 to produce a manifest at all
- [ ] #131, #132, #133 living-brief defects — **deferred by Fisher**, not blocking

---

# TRACK 1 — runnable now, no Apple identity

## 1A. Update signing key (free, ours)

Generated offline. Not an Apple credential.

- [ ] Generate the Ed25519 keypair locally. The private seed must never be printed, committed, or pasted into chat.
- [ ] Add secret `SOURCECADO_UPDATE_SIGNING_KEY` (base64 of the 32-byte seed)
- [ ] Add variable `SOURCECADO_UPDATE_KEY_ID`
[...]

PR #137, opened 2026-08-28
feat(desktop): tell the operator when a turn fails, and what kind of failure

### LINKED ISSUE #136: The interface drops the error message the server sends, so a failed turn looks like a freeze
Split out of #130. The provider half is fixed in #135. This is the reason that bug looked like a silent freeze instead of an error, and the reason its first diagnosis was wrong.

## Problem

When a turn fails, the server does the right thing. It emits an error event over the websocket and persists it to the session event log:

```json
{"version": 2, "type": "error", "session_id": "...", "run_id": "...",
 "state": "failed", "message": "The model provider failed after bounded recovery attempts."}
```

The interface renders nothing. The composer clears and the conversation sits silent. From the operator's chair a failed turn and a slow model are indistinguishable.

## Cause

`desktop/surfaces/gui/src/chat/store.ts:518`:

```typescript
if (event.type === "error") {
  this.apply({
    type: "message_state_changed",
    threadId: event.session_id,
    messageId: event.message_id,
    state: "failed",
  });
}
```

The `message` field is dropped. The store records that a message failed and discards the only text explaining why, so no surface has anything to show.

There is a second, separate gap. `isTransportError` (`store.ts:35`) requires `!("version" in event)`:

```typescript
if (typeof event !== "object" || event === null || "version" in event) {
  return false;
}
```

Server error events carry `"version": 2`, so they never satisfy that guard either. Both routes to showing an error miss.

## Why it matters

This is a trust problem, not a cosmetic one. An operator waiting on a reply cannot tell a dead conversation from a working one. During the #84 acceptance run this cost real time: two messages were sent into a permanently broken conversation and looked like they were being considered. The true cause was only found by connecting to the websocket directly and reading the raw traffic.

It also hides every other turn failure, not just #130. Rate limits, authentication problems, and configuration errors all produce specific, useful messages in `safe_provider_failure_message` (`desktop/coworker/provider_retry.py:363`). None of them reach the operator today.

## Fix

Carry `message` through into the failed message state and render it where the reply would have appeared. The wording is already operator-safe and deliberately free of provider internals, so it can be shown verbatim.

## Tests to add

- An error event with a message renders that message in the thread.
- A failed turn is visibly distinct from a running one.
- An error event carrying `version: 2` is handled, not silently dropped.
- The existing recoverable-notice path is unaffected.

FisherXZ:
Closes #136. Prerequisite for the Stage D checks in #134: a packaged QA pass is not worth running while the interface hides failures.

## Domain words

- **Terminal failure** — a turn that ended badly after retry and failover were already exhausted. There is no agent left to inform at that point; the model is the broken part.
- **Notice part** — an existing message part the client already renders as text. Used here rather than inventing a new surface.
- **error_kind** — a generic category (`rate_limit`, `authentication`, `provider`, `policy`, `internal`, ...). No provider internals.

## Problem

A failed turn rendered nothing. The composer cleared and the conversation went quiet.

The sidecar was doing its job: it emitted `{"type": "error", "state": "failed", "message": "..."}` over the websocket and wrote it to the session event log. `store.ts` handled the event by setting the message state to `failed` and discarding `message`, so no surface had anything to show.

This cost real time during the #84 run. Two messages were sent into a permanently broken conversation and looked like they were being considered. The true cause was only found by connecting to the websocket directly and reading raw frames, and the first diagnosis of #130 was wrong as a direct result.

## Before and after

| Situation | Before | After |
|---|---|---|
| Provider fails after retries | silence | the reason, rendered in the thread |
| Rate limited vs auth broken | indistinguishable | distinguishable by `error_kind` |
| Held turn (`outcome_unknown`) | own surface | unchanged, still its own surface |
| Older sidecar with no `error_kind` | n/a | parses, falls back to code `error` |

## Changes

**Sidecar.** All three `state: "failed"` sites now send `error_kind`, using the same value already passed to `turn_span.fail(...)`. The kind was computed and then dropped before it reached the client. Sending the span's own value means the operator view and the trace can never disagree.

The `held` event is untouched. It already carries `code: "outcome_unknown"` and an `effect_id`, and it is not a failure.

**Client.** The error branch keeps marking the message, and additionally adds a notice carrying the message text, keyed by `error_kind`. `addNotice` gained a `recoverable` flag so a terminal failure is not labelled recoverable; the existing recoverable-notice path passes `true` and is unchanged.

`error_kind` is optional in the protocol type, so an older sidecar still validates and falls back to code `error`.

## Why a notice rather than a new field

The client already has a notice part with `code`, `message`, and `recoverable`, and already renders notices as text. Reusing it means no new rendering path, no new adapter case, and no change to the message model. The whole client change is 23 lines.

## Measurements

```
1916 passed, 3 skipped, 1 warning in 79.08s
 Test Files  58 passed (58)
      Tests  519 passed (519)
tsc --noEmit   clean
```

Mutation harness, breaking one behaviour at a time and requiring the owning test to go red:

```
  red: server stops sending the kind on a terminal failure
  red: server stops sending the kind when no provider is configured
  red: client stops surfacing the failure text at all
  red: client marks a terminal failure as recoverable

4 red, 0 vacuous, 4 mutations
```

## Live verification

This branch is cut from `main`, where #130 is still unfixed, so a restarted sidecar produces a real terminal failure. Probed over a raw websocket, the persisted event now reads:

```json
{
  "version": 2,
  "type": "error",
  "state": "failed",
  "error_kind": "provider",
  "message": "The model provider failed after bounded recovery attempts."
}
```

`error_kind` is on the wire and on disk. Before this change that field did not exist.

## Context that shaped the scope

Fisher asked whether tool and environment errors should be fed back to the agent. They already are, and that is the right design: `_tool_message` (`turn.py:597`) returns error payloads to the model as tool results so it can adapt, and `had_tool_failure` makes the turn report `partial` rather than `ok` (`turn.py:1509`). Provider failures already retry three times per provider with jittered backoff, honour `Retry-After`, then fail over. None of that needed changing.

What was missing is only the last mile: once retry and failover are exhausted there is no agent to tell, and the human was told nothing.

## Still wrong, and not in this PR

**No retry affordance.** A failed turn still cannot be retried without retyping. #130 is the reason this is deliberate: a retry button on a permanently broken conversation would have been actively misleading. Retry UX should be gated on `error_kind`, which this PR makes possible but does not build.

**The generic message is still generic.** `safe_provider_failure_message` returns "The model provider failed after bounded recovery attempts." for anything unclassified. That text is now visible, which is the point, but it still does not name a cause. Improving the per-kind copy is separate.

**Only the three `failed` sites carry a kind.** `turn_end` states such as `partial` and `interrupted` do not, and are out of scope here.

### DIFF
--- a/desktop/coworker/turn.py
+++ b/desktop/coworker/turn.py
@@ -926,6 +926,7 @@ async def run_turn(
                 {
                     "type": "error",
                     "state": "failed",
+                    "error_kind": ErrorKind.POLICY.value,
                     "message": "This conversation's bound person file is unavailable.",
                 }
             )
@@ -938,6 +939,7 @@ async def run_turn(
             {
                 "type": "error",
                 "state": "failed",
+                "error_kind": ErrorKind.PROVIDER.value,
                 "message": "No model key. Set DEEPSEEK_API_KEY (deepseek-v4-pro) or MOONSHOT_API_KEY (kimi-k3) in ~/.config/club/.env.",
             }
         )
@@ -1712,6 +1714,7 @@ async def run_turn(
             {
                 "type": "error",
                 "state": "failed",
+                "error_kind": turn_error_kind.value,
                 "message": turn_error_message or str(exc),
             }
         )
--- a/desktop/coworker/update_channel/apply.py
+++ b/desktop/coworker/update_channel/apply.py
@@ -384,40 +384,65 @@ def sidecar_health_check(
     if not binary.is_file() or not os.access(binary, os.X_OK):
         return False
     token = secrets_module.token_hex(16)
-    port = _free_port()
     scratch = tempfile.mkdtemp(prefix="sourcecado-update-health-")
     environment = dict(os.environ)
     environment["CLUB_STATE_DIR"] = scratch
     environment["CLUB_API_TOKEN"] = token
     environment.pop("CLUB_EXIT_WITH_PARENT", None)
-    process = subprocess.Popen(
-        [str(binary), "--host", "127.0.0.1", "--port", str(port)],
-        cwd=tempfile.gettempdir(),
-        env=environment,
-        stdout=subprocess.DEVNULL,
-        stderr=subprocess.DEVNULL,
-    )
+    process = None
+    port = 0
     try:
         deadline = time.monotonic() + timeout
+        starts = 0
         while time.monotonic() < deadline:
+            if process is None:
+                if starts >= 3:
+                    return False
+                starts += 1
+                port = _free_port()
+                process = subprocess.Popen(
+                    [str(binary), "--host", "127.0.0.1", "--port", str(port)],
+                    cwd=tempfile.gettempdir(),
+                    env=environment,
+                    stdout=subprocess.DEVNULL,
+                    stderr=subprocess.DEVNULL,
+                )
             if process.poll() is not None:
-                return False
+                process.wait(timeout=1)
+                process = None
+                continue
+            connection = None
             try:
-                connection = http.client.HTTPConnection("127.0.0.1", port, timeout=2)
-                connection.request("GET", "/v1/health", headers={"X-Club-Token": token})
+                connection = http.client.HTTPConnection(
+                    "127.0.0.1", port, timeout=2
+                )
+                connection.request(
+                    "GET", "/v1/health", headers={"X-Club-Token": token}
+                )
                 response = connection.getresponse()
                 if response.status != 200:
-                    return False
+                    # The port we reserved was taken by something else.
+                    process.kill()
+                    try:
+                        process.wait(timeout=5)
+                    except subprocess.TimeoutExpired:
+                        pass
+                    process = None
+                    continue
                 return json.loads(response.read()).get("status") == "ok"
             except (OSError, ValueError):
                 time.sleep(0.25)
+            finally:
+                if connection is not None:
+                    connection.close()
         return False
     finally:
-        process.kill()
-        try:
-            process.wait(timeout=5)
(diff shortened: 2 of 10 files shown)


FisherXZ (commented):
## Verdict
**Fix, then ship.**

## Summary
A failed turn used to go quiet. The sidecar already sent an error event with operator-safe text. The client marked the message failed and dropped that text.

This PR sends `error_kind` from the three `state: "failed"` sites, using the same value the telemetry span already recorded. The client keeps the failed state and appends a notice with that text, marked not recoverable. Held turns stay on their own surface. An older sidecar without `error_kind` still parses.

The notice reuses the history-gap chrome. A rate-limited turn now reads as "Some conversation history is unavailable."

## Map
```mermaid
flowchart TD
  A[Operator sends a turn] --> B[Sidecar retries then fails over]
  B --> C[Error event with message and error_kind]
  C --> D[Store marks the assistant message failed]
  D --> E[Store appends a notice sibling]
  E --> F[Notice body is the sidecar text]
  E -.-> G[Title: Some conversation history is unavailable]
```

Why Fix, then ship: #136 exists so the operator can read why the turn died. The body text is in the thread. The title is a false history gap. The transport-error test already treated that title as a bug. Reusing the notice part is the right seam if the chrome matches the kind. It does not.

## Blocking
1. A terminal provider failure renders as "Some conversation history is unavailable." plus the real reason plus "Available messages are shown." I rendered ChatPage with `error_kind: "rate_limit"` and that is the string I got. #136 exists because silence caused a wrong first diagnosis of #130. This title is the same class of miss.

Issue #147, opened 2026-08-29
P0: Recover gracefully when the persisted active session no longer exists


FisherXZ:
## Parent

#134

QA reference: ISSUE-001 in the 2026-08-28 packaged preview report, build `0.0.1.212` at `9c20c26`.

## What to build

Make cold launch and quit/reopen reconcile the locally persisted active conversation with the sidecar's authoritative sessions before rendering chat home.

The packaged app currently requests a WKWebView-cached session ID that no longer exists, receives `404`, and renders “We couldn't load this conversation” with copy claiming the sidecar is unavailable even while the sidecar is healthy. Creating a new chat recovers, proving this is stale session selection rather than sidecar availability.

When the persisted session is gone, Sourcecado should automatically select another valid conversation or open a usable empty conversation. Existing person-bound chats and any recoverable draft must remain intact.

## Acceptance criteria

- [ ] Cold launch and quit/reopen are usable without pressing New chat or Reconnect when the persisted active session ID returns `404`.
- [ ] A missing active session selects a valid existing session or creates a usable empty conversation through one deterministic recovery path.
- [ ] The UI does not describe a healthy sidecar as unavailable when the failure is a missing session.
- [ ] Existing conversation rail entries, person-bound chats, and recoverable composer drafts survive recovery.
- [ ] Regression tests cover a stale persisted active ID on first paint and after an ordinary app restart.
- [ ] A packaged-preview smoke test proves the recovery path without a repository checkout or development virtual environment.

## Blocked by

None - can start immediately

FisherXZ:
Implementation is ready in #153. The GUI now waits for the authoritative session list before opening a cached chat, rejects stale saved chat IDs, falls back to a valid current conversation, and no longer blames a healthy sidecar for a missing conversation. Verification: 523 GUI tests passed and the production build passed. Packaged cold-launch/quit-reopen smoke remains after CI produces the artifact.

FisherXZ:
PR #153 CI is fully green: Desktop GUI, Python sidecar, and macOS preview artifact all passed; CodeRabbit completed with no actionable comments. The generated artifact is ready for the final cold-launch/quit-reopen smoke check.

FisherXZ:
[...]

PR #153, opened 2026-08-29
P0: Recover stale restored conversations

FisherXZ:
## Summary

- hold cached and WKWebView-retained chat destinations behind the authoritative first session refresh
- validate saved chat routes against fresh session IDs and fall back to a valid open or recent conversation
- recover a stranded draft in a fresh unbound conversation instead of copying person-specific text into another person's chat
- create a usable empty conversation automatically when no saved session survives
- stop telling the operator that a missing conversation means the sidecar is unavailable
- cover cold launch, ordinary reopen, draft survival, empty recovery, and missing-conversation copy with regressions

## Verification

- `npm test` — 58 files and 526 tests passed
- `npm run build` — TypeScript and Vite production build passed
- `git diff --check` — passed

## Follow-up acceptance

The packaged-preview cold-launch and quit/reopen smoke check remains under #134 once CI produces the updated artifact.

Addresses #147

### DIFF
--- a/desktop/surfaces/gui/src/app/AppShell.tsx
+++ b/desktop/surfaces/gui/src/app/AppShell.tsx
@@ -14,6 +14,7 @@ import {
 import { BoardView } from "../Board";
 import { PersonFileView } from "../PersonFile";
 import { ChatPage } from "../routes/ChatPage";
+import { moveDraft, readDraft } from "../chat/draftStorage";
 import { ConnectionsPage } from "../routes/ConnectionsPage";
 import { MemoryPage } from "../routes/MemoryPage";
 import { QuarantinePage } from "../routes/QuarantinePage";
@@ -30,6 +31,9 @@ import { readShellCache, writeShellCache } from "./sessionCache";
 
 export function AppShell() {
   const [cachedListing] = useState(() => readShellCache());
+  const cachedChatRestoreHash = cachedListing && isRestorableChatHash(window.location.hash)
+    ? window.location.hash
+    : null;
   const [route, setRoute] = useState(() => parseHash(window.location.hash));
   const [sessions, setSessions] = useState<SessionRow[] | null>(cachedListing?.sessions || null);
   const [openSessionId, setOpenSessionId] = useState<string | null>(cachedListing?.open_id || null);
@@ -46,8 +50,10 @@ export function AppShell() {
   const railFocusTimerRef = useRef<number | null>(null);
   const railWasOpenedRef = useRef(false);
   const searchReturnFocusRef = useRef<HTMLElement | null>(null);
-  const deferDestinationPersistenceRef = useRef(isRootHash(window.location.hash));
-  const cachedRestoreHashRef = useRef<string | null>(null);
+  const deferDestinationPersistenceRef = useRef(
+    isRootHash(window.location.hash) || Boolean(cachedChatRestoreHash),
+  );
+  const bootRestoreHashRef = useRef<string | null>(cachedChatRestoreHash);
   function openSearch() {
     searchReturnFocusRef.current =
       document.activeElement instanceof HTMLElement ? document.activeElement : null;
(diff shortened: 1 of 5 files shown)


FisherXZ (commented):
## Verdict
**Fix, then ship.**

## Summary
This PR holds a cached root restore behind the first authoritative session refresh. It validates the server's saved chat destination, falls back to an open or recent session, and removes the false sidecar-unavailable message. The existing 523-test GUI suite and production build pass. The root-hash cache case now works, but the ordinary WKWebView restart shape still loads the dead person chat. Recovery also drops the visible composer draft and stops at the welcome screen when no valid session remains.

## Map
```mermaid
flowchart TD
    A[Quit and reopen] --> B{Initial hash is root?}
    B -- Root --> C[Hold cached chat behind refresh]
    B -. Retained chat hash .-> D[Load missing person chat]
    D -. 404 remains after refresh .-> E[Manual conversation switch]
    C --> F[Fetch authoritative sessions]
    F --> G{Saved session still exists?}
    G -- Yes --> H[Open saved conversation]
    G -- No, valid session exists --> I[Open fallback conversation]
    I -. Draft stays under missing ID .-> J[Fallback composer is empty]
    G -- No sessions --> K[Show welcome screen]
    K -. Requires Start a chat .-> L[Usable empty conversation]
```

Why Fix, then ship: #147 requires cold launch and ordinary restart to recover through one deterministic path. It also requires recoverable drafts to survive and an empty conversation when no valid session exists. I added one focused regression for each criterion against this head; all three failed.

## Blocking
1. When WKWebView reopens on `#/chat/missing/person/person-old`, Sourcecado requests that missing conversation before the session list resolves and remains on the 404 card after the fresh `alpha` session appears. This is the ordinary quit/reopen path that #147 exists to fix.
2. When root restoration correctly falls back from missing session `missing` to valid session `alpha`, the unsent draft remains keyed only to `missing`; the visible fallback composer is empty. This violates the criterion that a recoverable composer draft survives recovery.
3. When the authoritative list contains no surviving session, Sourcecado shows the welcome screen and never calls `createSession`; the operator must press Start a chat. This violates the automatic valid-session-or-usable-empty-conversation recovery criterion.

FisherXZ:
Addressed all three manual-review blockers in 62cc1e2: retained WKWebView hashes now wait for fresh reconciliation; stale person-chat drafts recover into a new unbound conversation so they cannot leak into another person chat; and an empty authoritative list creates a usable conversation automatically. Focused blocker regressions pass, along with 526 GUI tests and the production build.

FisherXZ (commented):
## Verdict
**Ship.**

## Summary
Commit `62cc1e2` closes all three blockers from the first review. A WKWebView-retained chat hash now enters the same boot reconciliation gate as a root restore, so the missing conversation is never requested before fresh sessions arrive. A stranded draft creates a new unbound conversation and moves there instead of entering another person's valid chat. An authoritative empty list also creates and opens a usable conversation automatically. The three focused regressions, all 526 GUI tests, and the production build pass.

## Map
```mermaid
flowchart TD
    A[Cold launch or quit/reopen] --> B[Hold restored chat behind boot gate]
    B --> C[Fetch authoritative sessions]
    C --> D{Saved session survives?}
    D -- Yes --> E[Open saved conversation]
    D -- No, no draft --> F{Another session exists?}
    F -- Yes --> G[Open valid fallback]
    F -- No --> H[Create usable empty conversation]
    D -- No, draft exists --> I[Create new unbound conversation]
    I --> J[Move draft to new session]
    J --> K[Open recovered composer]
```

Why Ship: The restart path now reconciles before any transcript request. Draft recovery preserves the operator's text without crossing a person boundary. The empty-state path finishes automatically.

## Blocking
None.

The failure I was most afraid of was a missing person-chat draft being copied into another person's fallback conversation. The focused regression opens a fresh unbound `recovered-draft` chat, keeps `alpha` unloaded, and renders `Unsent note` only in the new composer.

